PR #9, opened 2019-06-18
Fix paths and add copyright boilerplate

jcrossley3:
Minor fixes resulting from the move


jchesterpivotal:
I would make a joke here about /symboliclgtm, but then I realised that's why github added emoji.

test/e2e/knative_serving_test.go
```diff
@@ -4,8 +4,8 @@ import (
 	"testing"
```
houshengbo: We do not have header in this file.

jcrossley3: Good catch!

PR #12, opened 2019-06-18
Add update-deps.sh

houshengbo:
The changes in this PR is the result of running ./hack/update-deps.sh for the first time.


hack/update-deps.sh
```diff
@@ -0,0 +1,32 @@
...
+cd ${REPO_ROOT_DIR}
+
+# Ensure we have everything we need under vendor/
+dep ensure
+
+rm -rf $(find vendor/ -name 'OWNERS')
+rm -rf $(find vendor/ -name '*_test.go')
+rm -fr vendor/github.com/knative/test-infra/devstats
+
+remove_broken_symlinks ./vendor
```
k4leung4: should we also call update_licenses similar to the knative/serving repo?
https://github.com/knative/serving/blob/master/hack/update-deps.sh#L36

houshengbo: I am not sure we need it for now, since we do not have fir third_party

Gopkg.lock
```diff
@@ -475,6 +474,23 @@
   revision = "4b7aa43c6742a2c18fdef89dd197aaae7dac7ccd"
   version = "1.0.1"
 
+[[projects]]
+  digest = "1:7226c7e52dcd9273715f24d6e3f023857ccfc955c2bb09fcd9fea733b6cbb96f"
+  name = "github.com/openshift-knative/knative-serving-operator"
```
jcrossley3: We don't want this, since this was the old repo location.

houshengbo: OK

hack/update-deps.sh
k4leung4: works for me.

PR #14, opened 2019-06-19
Parameterize release script to fix #6

### LINKED ISSUE #6: Operator should be released to gcr.io/knative-releases/github.com/knative
Currently the ./hack/release.sh deploys to quay.io.

jcrossley3:
These values reflect what we currently do for quay.io, but the
defaults should really be changed to what's required for their new
target location on gcr.io.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -14,18 +14,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Usage: REGISTRY=xxx ORG=yyy IMAGE=zzz ./hack/release.sh [--tag]
+
 set -ex
 
 pushd $(dirname "$0")/..
 
-readonly HUB_ORG="${HUB_ORG:-"openshift-knative"}"
-
-IMAGE=$(basename $(pwd))
+REGISTRY=${REGISTRY:-quay.io}
+ORG=${ORG:-openshift-knative}
+IMAGE=${IMAGE:-knative-serving-operator}
 
 eval VERSION=v$(grep Version version/version.go | awk '{print $3}')
-operator-sdk build quay.io/$HUB_ORG/$IMAGE:$VERSION
-docker push quay.io/$HUB_ORG/$IMAGE:$VERSION
-git tag -f $VERSION
-git push --tags --force
+
+TARGET=$REGISTRY/$ORG/$IMAGE:$VERSION
+
+operator-sdk build $TARGET
+docker push $TARGET
+
+if [ "$1" == "--tag" ]; then
+  git tag -f $VERSION
+  git push --tags --force
+fi
 
 popd


trshafer:
Should this link to test-infra/scripts/release in the way knative serving does? https://github.com/knative/serving/blob/b55af78dc41d36fe0ebfb94514826fa9647aaa87/hack/release.sh#L17

jcrossley3:
> Should this link to test-infra/scripts/release in the way knative serving does? https://github.com/knative/serving/blob/b55af78dc41d36fe0ebfb94514826fa9647aaa87/hack/release.sh#L17

I don't know. It's entirely possible that the script we used to push to quay.io is completely irrelevant in light of knative's release tooling/process. I would yield to whoever has experience with that process.

k4leung4:
@dushyanthsc Can you comment on what should be appropriate to integrate with the knative release process, thanks

dushyanthsc:
/cc: @nbarthwal who owns Knative serving releases. Can you please help answer the question?

nbarthwal:
If the goal is to use the test-infra tooling to release, then yes - think link https://github.com/knative/serving/blob/b55af78dc41d36fe0ebfb94514826fa9647aaa87/hack/release.sh#L17 does it write. Just implement the build_release and source the script. There is another step - the autorelease? which is a prow job that runs every 30 mins and release if there is a branch of the release-X.X for which the tag does not exists. Is there somerthing you need? For that, the prow config files need to add extra stuff.

hack/release.sh
```diff
@@ -14,18 +14,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Usage: REGISTRY=xxx ORG=yyy IMAGE=zzz ./hack/release.sh [--tag]
+
 set -ex
```
houshengbo: @jcrossley3 Jim, I am thinking of whether we should write release.sh the same way as in https://github.com/knative/serving/blob/master/hack/release.sh.

Images will be published under gcr.io/knative-releases/ as released artifacts.
Nightly build is put under: gcr.io/knative-nightly/, if necessary.

As long as we do source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/release.sh. These params have been used automatically.  The only thing we need to implement is a function called build_release.

houshengbo:
@nbarthwal  Since this repo is new, do we need to change the yaml in test-infra in order to make sure the prow job can do release check every 30 mins?  I had a PR for test-infra, not sure whether that is enough: https://github.com/knative/test-infra/pull/973.

jcrossley3:
I'm gonna close this since it's not relevant to the test-infra tooling. @houshengbo will create a new PR once the repo has been adapted to `ko`

nbarthwal:
> @nbarthwal @jcrossley3 I had a PR for test-infra, not sure whether that is enough: [knative/test-infra#973](https://github.com/knative/test-infra/pull/973). I think I have enabled the unit, integration, build tests, and release. Could you review it? Thx.

No, you need auto release jobs added. See https://github.com/nbarthwal/test-infra/blob/bf22e8abd4b50ce60cc5ba87083ab0347c45e8f9/ci/prow/config.yaml#L1844 for example

PR #20, opened 2019-06-20
Add ability to update the image and tag of knative serving images

### LINKED ISSUE #1: Operator should be able to install from different registries and tags
Currently knative serving release uses the registry: `gcr.io/knative-releases` and the build path: `github.com/knative/serving/cmd`. Additionally, the tag is built using a sha `sha256:1e40c99ff5977daa2d69873fff604c6d09651af1f9ff15aadf8849b3ee77ab45` instead of a tag number `0.6.0`. These fields should be customizable, so it could point to a registry of `quay.io/openshift-knative` or others.

I think it's reasonable to have reasonable defaults from knative serving, but support modifications by changing `KnativeServing` CR.

The configuration could look like:
```YAML
kind: KnativeServing
metadata:
  name: knative-serving
spec:
  config:
  image:
    registry: "my-registry.io/my-knative-release/path"
    tag: "v0.6.0"
```

trshafer:
Fixes #1.


pkg/controller/knativeserving/common/extensions.go
```diff
@@ -63,6 +64,14 @@ func (exts Extensions) Transform(instance *servingv1alpha1.KnativeServing) []mf.
 				UpdateConfigMap(u, data, log)
 			}
 		}
+		if u.GetKind() == "Deployment" {
```
greghaynes: The comment on line 60 should get moved in to this function block now since it is not purely about config overriding

pkg/controller/knativeserving/common/helpers.go
```diff
@@ -34,3 +40,55 @@ func UpdateConfigMap(cm *unstructured.Unstructured, data map[string]string, log
 		unstructured.SetNestedField(cm.Object, v, "data", k)
 	}
 }
+
+// Set some data in a configmap, only overwriting common keys if they differ
```
greghaynes: Is this comment leftover from somewhere?

I'd also expect it to start with // UpdateImage

pkg/controller/knativeserving/common/extensions.go
trshafer: Done

pkg/controller/knativeserving/common/helpers.go
trshafer: Done

pkg/controller/knativeserving/common/extensions.go
```diff
@@ -56,8 +57,17 @@ func (exts Extensions) Transform(instance *servingv1alpha1.KnativeServing) []mf.
...
 		result = append(result, extension.Transformers...)
 	}
-	// Let any config in instance override everything else
 	return append(result, func(u *unstructured.Unstructured) error {
+		// Update the deployment with the new registry and tag
+		if u.GetKind() == "Deployment" {
+			registry := instance.Spec.Image.Image
+			tag := instance.Spec.Image.Tag
+			var deployment *appsv1.Deployment
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, deployment); err == nil {
```
jcrossley3: I've been using `scheme.Convert` to do this in other places, but I'm not sure if either is preferred. It makes me think maybe the controller's scheme should be made available in this function though.

pkg/controller/knativeserving/common/extensions.go
```diff
@@ -56,8 +57,17 @@ func (exts Extensions) Transform(instance *servingv1alpha1.KnativeServing) []mf.
 	for _, extension := range exts {
 		result = append(result, extension.Transformers...)
 	}
-	// Let any config in instance override everything else
 	return append(result, func(u *unstructured.Unstructured) error {
```
jcrossley3: If we start adding more conditions in here, it might make more sense to define `mf.Transformer` functions elsewhere in the file and reference them here, similar to the `mf.Inject*` functions above (though they wouldn't have to be closures). I'll let you decide whether we do that now or later.

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -38,6 +45,8 @@ type KnativeServingSpec struct {
 	// A means to override the corresponding entries in the upstream configmaps
 	// +optional
 	Config map[string]map[string]string `json:"config,omitempty"`
+
+	Image Registry `json:"registry,omitempty"`
```
jcrossley3: I think we should include the proper updates to `deploy/crds/serving_v1alpha1_knativeserving_crd.yaml` with this PR. Either manually or via `operator-sdk generate k8s`

pkg/controller/knativeserving/common/helpers.go
```diff
@@ -34,3 +41,93 @@ func UpdateConfigMap(cm *unstructured.Unstructured, data map[string]string, log
 		unstructured.SetNestedField(cm.Object, v, "data", k)
 	}
 }
+
```
jcrossley3: I don't love the name "helpers.go" for this file (my fault), and you've added so many related functions that I might lean toward putting them in their own say, "images.go" file? Still in the `common` package, of course.

trshafer: Done

pkg/apis/serving/v1alpha1/knativeserving_types.go
trshafer: Done

pkg/controller/knativeserving/common/extensions.go
trshafer: Done

trshafer:
I have verified that this works with:

```
kind: KnativeServing
metadata:
  name: knative-serving
spec:
  registry:
    default-registry: "my-registry.io/my-build/knative/cmd"
    default-tag: "v0.6.0"
```

How do you feel about this CRD?

config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -51,6 +51,16 @@ spec:
               description: A means to override the corresponding entries in the upstream
                 configmaps
               type: object
+            registry:
+              properties:
+                desfult-registry:
```
jcrossley3: typo

config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -51,6 +51,16 @@ spec:
...
+                  description: The default registry to use for all knative images.
+                  type: string
+                desfult-tag:
```
jcrossley3: ditto

pkg/controller/knativeserving/common/extensions.go
```diff
@@ -67,6 +73,26 @@ func (exts Extensions) Transform(instance *servingv1alpha1.KnativeServing) []mf.
 	})
 }
 
+func updateDeployment(scheme *runtime.Scheme, instance *servingv1alpha1.KnativeServing, u *unstructured.Unstructured) error {
+
+	var deployment = &appsv1.Deployment{}
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, deployment)
```
jcrossley3: is this better than `scheme.Convert(u, deployment)`?

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -28,6 +28,13 @@ const (
 	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
 )
 
+// Registry defines the registry of KnativeServing
+// +k8s:openapi-gen=true
+type Registry struct {
+	DefaultRegistry string `json:"default-registry,omitempty"`
+	DefaultTag      string `json:"default-tag,omitempty"`
+}
```
jcrossley3: What is the "Default*" prefix meant to imply? Why not just `Name` and `Tag` as the only elements of the `Registry` struct?

jcrossley3:
> How do you feel about this CRD?
> 
> Should it be:
> 
> ```yaml
> spec:
>   registry:
>     default:
>       registry: my-registry.io/my-build/knative/cmd
>       tag: v0.6.0
> ```

Personally, I'd prefer not to see "registry" used in two places: how about just `name` and `tag`?

> It's a bit strange to have a `default` but not support the other services in the crd. Should this also support the other services before merging?

What other services? Like activator or autoscaler? So you could say use this registry for everything except autoscaler? And it could come from a different registry?

jcrossley3:
Also, would we expect this to be affected as well? https://github.com/knative/serving-operator/blob/6c344b7842fb589c2993ebf494e48238537645f6/config/resources/knative-serving-0.6.0.yaml#L997

pkg/controller/knativeserving/common/images_test.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+var updateImageTests = []updateImageTest{
+	{
+		name:     "NewRegistryAndTagFromDigest",
+		in:       "gcr.io/knative-releases/github.com/knative/serving/cmd/queue@sha256:1e40c99ff5977daa2d69873fff604c6d09651af1f9ff15aadf8849b3ee77ab45",
+		registry: "new-registry.io/test/path",
+		tag:      "new-tag",
+		expected: "new-registry.io/test/path/queue:new-tag",
+	},
```
jcrossley3: Another challenge just occurred to me: RHT currently renames its images with a `knative-serving-` prefix, so `queue` becomes `knative-serving-queue`. Hence, our image name will be `quay.io/openshift-knative/knative-serving-queue:v0.6.0`. Will the current logic stick a `/` in there? If so, can we just expect the separator to be in the name, e.g `quay.io/openshift-knative/knative-serving-` and `new-registry.io/test/path/`?

config/crds/serving_v1alpha1_knativeserving_crd.yaml
trshafer: done

config/crds/serving_v1alpha1_knativeserving_crd.yaml
trshafer: done

trshafer:
Per comment https://github.com/knative/serving-operator/pull/20#issuecomment-505664419, @greghaynes suggested an option that supports both a default and an individual override: https://github.com/knative/serving-operator/issues/1#issuecomment-500942893. Your suggestion is the correct assumption. I should probably the ability to override the default per service in this PR, so I don't need to circle back again.

pkg/apis/serving/v1alpha1/knativeserving_types.go
trshafer: Per comment https://github.com/knative/serving-operator/pull/20#issuecomment-505933467, it demonstrates the that this is the default registry and tag to use. I am starting to prefer the nested default map.

trshafer:
Per comment https://github.com/knative/serving-operator/pull/20#issuecomment-505870232, that's a good question. I don't think so, but it would be surprising. Is it unreasonable for the user to make the CR:

```yaml
spec:
  config:
    deployment: my-registry.io/my-build/knative/cmd/queue:v0.6.0
  registry:
    default:
      name: my-registry.io/my-build/knative/cmd
      tag: v0.6.0
```

What's unfortunate is the user will always need to do this, which is pretty non-obvious. @greghaynes @jcrossley3 what do you think?

trshafer:
I'm leaning towards Option 1.

### Option 1 
```yaml
spec:
  registry:
    default:
      name: my-registry.io/my-build/knative/cmd
      tag: v0.6.0
    autoscaler: 
      image: my-registry.io/my-build/knative/cmd/autoscaler:v0.6.1
```
This option follows a similar pattern to the config override. 

### Option 2
```yaml
spec:
  registry:
    default:
      name: my-registry.io/my-build/knative/cmd
      tag: v0.6.0
    autoscaler: my-registry.io/my-build/knative/cmd/autoscaler:v0.6.1
```
The disadvantage of this option is that there are two implicit keys. The autoscaler key is both the name of the service and the key indicates the image location. It would be impossible to add another key to the override. I don't know what that key would be right now, but I think this format puts us into a corner.

### Option 3
```yaml
spec:
  registry:
    default:
      name: my-registry.io/my-build/knative/cmd
      tag: v0.6.0
  overrides:
    - name: autoscaler
      image: my-registry.io/my-build/knative/cmd/autoscaler:v0.6.1
    - name: controller
      image: my-registry.io/my-build/knative/cmd/controller:v0.6.1a
```

This option is the most verbose and most flexible. There are no implicit keys. The keys are purely for structure and do not carry additional meaning. However I think the user experience suffers.

jcrossley3:
> I'm leaning towards Option 1.

Of those, I lean toward Option 1 as well. How do you feel about something like this? 

### Option 4
```yaml
spec:
  registry:
    default: quay.io/openshift-knative/knative-serving-${NAME}:v0.6.0
    autoscaler: my-registry.io/my-build/knative/cmd/autoscaler:v0.6.1
```
It helps account for the path separator problem I mentioned in a previous comment due to our weird RHT naming scheme for the images we build from our downstream forks.

trshafer:
WRT https://github.com/knative/serving-operator/pull/20#issuecomment-505955152, I like this. I dramatically simplifies the registry configuration. I wish there were a way in Go to specify "default" as a special key in the map, but I don't think there is.

jcrossley3:
A nit, but maybe `override` instead of `images`? https://github.com/knative/serving-operator/pull/20#issuecomment-505957287

pkg/controller/knativeserving/common/extensions.go
trshafer: I couldn't get `scheme.Convert(u, deployment)` to work. I got error message `TypeMeta not present in src` which comes from https://github.com/kubernetes/apimachinery/blob/961b39a1baa06f6c52bdd048a809b9f5b47f1337/pkg/conversion/converter.go#L828. The issue I believe is `TypeMeta` https://github.com/knative/serving-operator/blob/6c0272eca46470454567e78aafbfd53cf7a8f53c/vendor/k8s.io/api/apps/v1/types.go#L255 is not in the unstructured map. It's unable to take the inline json content and convert it to a Deployment. This way worked. 

I'll look more into the scheme tests and see if I can reproduce this issue: https://github.com/kubernetes/apimachinery/blob/961b39a1baa06f6c52bdd048a809b9f5b47f1337/pkg/runtime/scheme_test.go#L1

greghaynes:
I really like the idea of breaking out each image url as a separate property. In this setup we shouldn't need any kind of string templating right? IMO forcing operators to use specific image names in their registry is an anti-feature caused by templating ${NAME}.

greghaynes:
OK just chatted with @trshafer and the templating in default makes sense to me now :)

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -28,6 +28,15 @@ const (
 	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
 )
 
+// Registry defines image overrides of knative images.
+// The default value is used as a default format to override for all knative deployments.
+// The override values are specific to each knative deployment.
+// +k8s:openapi-gen=true
+type Registry struct {
+	Default  string            `json:"default,omitempty"`
```
greghaynes: Can we add comments to both of these properties explaining default vs override? Being particular about this because this is what will end up in API docs (which is what users will see).

trshafer: Done.

greghaynes:
/lgtm

Sorry it took so many review cycles!

trshafer:
Hey @greghaynes could you approve as well? Thanks.

config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -51,6 +51,19 @@ spec:
               description: A means to override the corresponding entries in the upstream
                 configmaps
               type: object
+            registry:
+              description: A means to override the registry for all knative images.
+              type: object
+              properties:
+                default:
+                  description: The default registry to use for all knative iamges.
```
evankanderson: Is this a registry or an image reference template?

pkg/controller/knativeserving/common/images.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// UpdateImage updates the image of the deployment with a new registry and tag
+// @param image takes the form of new-registry.io/full/path/${NAME}:new-tag
+func UpdateImage(deployment *appsv1.Deployment, registry *servingv1alpha1.Registry, log logr.Logger) {
+	image := imageName(deployment.GetName(), registry)
+	if image == "" {
+		return
+	}
+	image = strings.ReplaceAll(image, "${NAME}", deployment.GetName())
```
evankanderson: Did you want `${NAME}` to work in `overrides` as well as `default`? If not, you may want the `${NAME}` substitution to happen in the `imageName` function (but see below).

pkg/controller/knativeserving/common/images.go
```diff
@@ -0,0 +1,57 @@
...
+	containers := deployment.Spec.Template.Spec.Containers
+	for index := range containers {
+		container := &containers[index]
+		updateContainer(container, image, log)
```
evankanderson: This formulation means that any Pod that has more than one container will run the same image for all containers. That's probably not what is desired -- maybe you want to use the name of the container, rather than the name of the Deployment?

pkg/controller/knativeserving/common/images.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	log.V(1).Info("Finished updating images", "deployment", deployment.GetName())
+}
+
+func imageName(name string, registry *servingv1alpha1.Registry) string {
+	defaultRegistry := registry.Default
```
evankanderson: I'd find it slightly more idiomatic to avoid the temporary here, and just `return registry.Default` at the end.

pkg/controller/knativeserving/common/images.go
```diff
@@ -0,0 +1,57 @@
...
+	overrideImage := registry.Override[name]
+	if overrideImage != "" {
+		return overrideImage
+	}
+	return defaultRegistry
+}
+
+func updateContainer(container *corev1.Container, image string, log logr.Logger) {
+	oldImageName := container.Image
```
evankanderson: Again, I don't think this temporary adds any clarity in a function of this length.

config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -51,6 +51,19 @@ spec:
...
+                    Takes the form of example-registry.io/custom/path/${NAME}:custom-tag
+                  type: string
+                override:
+                  description: The full image location of an individual knative service.
```
evankanderson: I'm not sure you want to use the term "service" here. Maybe "container" if you follow my suggestion below?

pkg/controller/knativeserving/common/images.go
trshafer: Done.

pkg/controller/knativeserving/common/images.go
trshafer: Done.

pkg/controller/knativeserving/common/images.go
trshafer: Done.

pkg/controller/knativeserving/common/images.go
trshafer: You bring up a good point that this string will override all containers. We have a few options. We can either put the ${NAME} interpolation back into the override which will use the container name instead of the deployment name. Or we can have individual overrides per container. I think the interpolation is easier to understand and consume, however it makes it more confusing or unusable if the images are pretty unrelated and now need to follow the same registry and path.

jcrossley3: Personally, I think the keys in the override map should be container names rather than deployment names, since only containers have images.
That being said, these transforms will only apply to the upstream manifest packaged in kodata. And as of 0.7.0, all deployments have a single container and their names are identical. If we expect that to change in a subsequent release, we should absolutely account for it now. But if we don't, I'm not sure adding the complexity now is worth it.

jcrossley3: Do we expect deployments A and B to each have a container C anytime soon? Or is it sufficient to expect a naming convention of A-C and B-C for those containers? Personally, I think the Default is going to account for 99% of the usecases. I suppose we could consider the Override to be a map of maps, with the first key being the Deployment name and the second being the Container name?

trshafer: I agree. I think the map of maps is likely excessive right now. It would add a lot of additional complexity and current duplication:

```yaml
overrides:
  activator:
    activator: "new-activator-image"
  controller:
    controller: "new-controller-image"
```

However if one of the containers required a security fix in a deployment with multiple containers, the tags will need to be updated on both those containers which is non-obvious. This is especially concerning as the user would need to know this while addressing a potential production issue.

jcrossley3: But as long as all containers are uniquely named within the entire manifest, it's not an issue, correct?
And again, you guys are worrying me by suggesting we might have a deployment with multiple containers? Why do you expect that to happen? We control the manifest, not the user.

greghaynes: I do like the idea of using container names rather than deployment names here - should be relatively simple to implement and somewhat future proofs us if we add additional containers to a deployment in serving sometime. We probably want a static whitelist as well though of containers we expect to see. I dont think we need to worry too much about the nuances of what happens when serving runs multiple containers and someone wants to roll out a security fix etc - we're targeting a known quantity at this point and can worry about that if/when serving creates that case.

evankanderson: I think using container name here should be fine, assuming that we communicate to the serving team that expectation. I worry that not designing for Pods with more than one container will probably result in a spectacular failure at some future point, but I think enforcing that container names are a unique mapping with images is probably reasonable.

**Specific scenario I'm thinking about:**
Imagine there is some sort of "helper" container (e.g. fluentd for copying logs) that we want to run alongside several components. I think using the same container name might be convenient here to allow updating the helper across several different deployments.

config/crds/serving_v1alpha1_knativeserving_crd.yaml
trshafer: Image reference template. Updated the comment.

pkg/controller/knativeserving/common/images.go
trshafer: Updated the code to use container names instead of the deployment name.

jcrossley3: > **Specific scenario I'm thinking about:**
> Imagine there is some sort of "helper" container (e.g. fluentd for copying logs) that we want to run alongside several components. I think using the same container name might be convenient here to allow updating the helper across several different deployments.

In that scenario, I would expect all of those fluentd containers across all those deployments to use the same image, so letting them all have the same name should be fine, I think. It's only the like-named containers that serve different purposes where i think the container-keyed override design might break down, but I'm not convinced that's a very likely scenario.

pkg/controller/knativeserving/common/extensions.go
vincent-pli: I think it's one challenge here, just consider the image url in `deployment`. 
Actually, there is image url in `configmap`, event in `args` in `pod` definition.
Could we have a common solution to pick them up and do replacement, use `regex`  or something.

pkg/controller/knativeserving/common/extensions.go
vincent-pli: Right, if `manifestival` could supply something like `InjectImageURL`, that's will be great.

pkg/controller/knativeserving/common/extensions.go
trshafer: I'm concerned with a common solution that supports replacing all possible values. I think if the configmap values need to be changed, the use should use the `config` key that mutates configmaps.

The `Image` kind needs to be able to be updated though.

trshafer: Updated to support `Image`

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -28,6 +28,23 @@ const (
...
 
+// Registry defines image overrides of knative images.
+// This affects both apps/v1.Deployment and caching.internal.knative.dev/v1alpha1.Image.
+// The default value is used as a default format to override for all knative deployments.
+// The override values are specific to each knative deployment.
+// +k8s:openapi-gen=true
+type Registry struct {
+	// The default image reference template to use for all knative images.
+	// It takes the form of example-registry.io/custom/path/${NAME}:custom-tag
+	// ${NAME} will be replaced by the deployment name.
```
jcrossley3: Deployment name? Container? Image? Basename?

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -38,6 +55,11 @@ type KnativeServingSpec struct {
 	// A means to override the corresponding entries in the upstream configmaps
 	// +optional
 	Config map[string]map[string]string `json:"config,omitempty"`
+
+	// A means to override the corresponding deployment images in the upstream deployments.
```
jcrossley3: `s/deployment //`

pkg/controller/knativeserving/common/extensions.go
```diff
@@ -67,6 +77,40 @@ func (exts Extensions) Transform(instance *servingv1alpha1.KnativeServing) []mf.
 	})
 }
 
+func updateCachingImage(scheme *runtime.Scheme, instance *servingv1alpha1.KnativeServing, u *unstructured.Unstructured) error {
+	var image = &caching.Image{}
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, image)
```
jcrossley3: I'll look into why `scheme.Convert` doesn't work after this is merged.

pkg/apis/serving/v1alpha1/knativeserving_types.go
trshafer: Done

pkg/apis/serving/v1alpha1/knativeserving_types.go
trshafer: Done

PR #22, opened 2019-06-21
Remove comand from operator container

greghaynes:
This is overriding the exec reference ko creates


trshafer:
Does this still have the runtime error, or will that be fixed in a subsequent PR?

PR #29, opened 2019-06-21
Temporarily stop waiting for knative-serving until tests work

trshafer:


greghaynes:
uhoh - looks like that success command doesn't exist without sourcing test-infra

PR #42, opened 2019-07-01
Fix update-codegen to generate clients

greghaynes:
We need this for using pkg/controller. When update-codegen was copied in
we seem to have left the paths pointing to knative/serving instead of
serving-operator.


pkg/apis/serving/v1alpha1/register.go
```diff
@@ -25,10 +25,17 @@ import (
...
+}
+
 var (
 	// SchemeGroupVersion is group version used to register these objects
 	SchemeGroupVersion = schema.GroupVersion{Group: "serving.knative.dev", Version: "v1alpha1"}
 
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
 	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+
+	AddToScheme = SchemeBuilder.AddToScheme
```
houshengbo: I suggest we have the standard function as below:
```
// addKnownTypes adds the set of types defined in this package to the supplied
// scheme.
func addKnownTypes(s *runtime.Scheme) error {
    s.AddKnownTypes(SchemeGroupVersion,
        &KnativeServing{},
        &KnativeServingList{})
    metav1.AddToGroupVersion(s, SchemeGroupVersion)
    return nil
}
```

Change the SchemeBuilder into SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes).
However the old SchemeBuilder should be kept as well, since we are transitioning.
e.g. SchemeBuilderSDK  = &scheme.Builder{GroupVersion: SchemeGroupVersion}

houshengbo:
Thx, Greg.

PR #44, opened 2019-07-01
Update operator to install knative serving 0.7.0

trshafer:
I validated this by deploying the old version of 0.6.0 using the operator and then deploying 0.7.0 using the operator on GKE.

I did not delete the 0.6.0 yaml files.

### DIFF
--- a/cmd/manager/kodata/knative-serving/0.7.0.yaml
+++ b/cmd/manager/kodata/knative-serving/0.7.0.yaml
@@ -3,7 +3,7 @@ kind: Namespace
 metadata:
   labels:
     istio-injection: enabled
-    serving.knative.dev/release: "v0.6.0"
+    serving.knative.dev/release: "v0.7.0"
   name: knative-serving
 
 ---
@@ -13,7 +13,7 @@ metadata:
   labels:
     networking.knative.dev/certificate-provider: cert-manager
     serving.knative.dev/controller: "true"
-    serving.knative.dev/release: "v0.6.0"
+    serving.knative.dev/release: "v0.7.0"
   name: knative-serving-certmanager
 rules:
 - apiGroups:
@@ -36,7 +36,7 @@ metadata:
   labels:
     networking.knative.dev/ingress-provider: istio
     serving.knative.dev/controller: "true"
-    serving.knative.dev/release: "v0.6.0"
+    serving.knative.dev/release: "v0.7.0"
   name: knative-serving-istio
 rules:
 - apiGroups:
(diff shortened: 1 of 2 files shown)


cmd/manager/kodata/knative-serving-0.7.0.yaml
```diff
@@ -0,0 +1 @@
+../../../config/resources/knative-serving-0.7.0.yaml
```
jcrossley3: I'd rather this symlink not exist.

config/resources/knative-serving-0.7.0.yaml
```diff
@@ -0,0 +1,1851 @@
+apiVersion: v1
```
jcrossley3: I'd prefer this file to reside beneath `cmd/manager/kodata/knative-serving/0.7.0.yaml`. The original choice to put it in `deploy/resources` (since renamed `config/resources`) was totally arbitrary, and I'd rather we not deal with symlinks just to preserve that arbitrary choice. :)

pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -105,7 +105,7 @@ type ReconcileKnativeServing struct {
 // Create manifestival resources and KnativeServing, if necessary
 func (r *ReconcileKnativeServing) InjectClient(c client.Client) error {
 	koDataDir := os.Getenv("KO_DATA_PATH")
-	m, err := mf.NewManifest(filepath.Join(koDataDir, "knative-serving-0.6.0.yaml"), *recursive, c)
+	m, err := mf.NewManifest(filepath.Join(koDataDir, "knative-serving-0.7.0.yaml"), *recursive, c)
```
jcrossley3: I don't like having to update this file for every release. We can just agree on a directory that will always contain the correct version of knative-serving, and manifestival, like kubectl, will apply all the manifests it finds beneath that directory (typically just one):
```suggestion
	m, err := mf.NewManifest(filepath.Join(koDataDir, "knative-serving/"), false, c)
```

trshafer:
@jcrossley3 good idea. I separated out the requested change to a separate PR: https://github.com/knative/serving-operator/pull/45

trshafer:
/hold

until https://github.com/knative/serving-operator/pull/45 is merged

trshafer:
/unhold

#45 is merged

@jcrossley3 I separated out the change to the structure from the update to 0.7.0. Should this also update the version of the operator?

trshafer:
/hold cancel

#45 is merged

@jcrossley3 I separated out the change to the structure from the update to 0.7.0. Should this also update the version of the operator?

PR #46, opened 2019-07-02
Switch to use knative.dev/pkg instead of github.com/knative/pkg

houshengbo:
In order to keep consistent with all the knative projects, we replace package github.com/knative/pkg with knative.dev/pkg.


Gopkg.toml
```diff
@@ -91,6 +95,7 @@ required = [
   name = "github.com/knative/test-infra"
   non-go = false
 
-[[constraint]]
-  branch = "master"
-  name = "github.com/knative/pkg"
+[[prune.project]]
+  name = "knative.dev/pkg"
+  unused-packages = false
+  non-go = false
```
mattmoor-sockpuppet:
```suggestion
  non-go = false

```

Gopkg.toml
```diff
@@ -91,6 +95,6 @@ required = [
   name = "github.com/knative/test-infra"
   non-go = false
 
-[[constraint]]
-  branch = "master"
-  name = "github.com/knative/pkg"
+[[prune.project]]
```
greghaynes: Why do we need to add this prune?

houshengbo: We only keep the packages when they are used.

houshengbo: In general, we prune go-tests, non-go and unused-packages. For knative.dev/pkg, we keep non-go, e.g. the license.

greghaynes: sounds good

greghaynes:
When I run hack/update-deps.sh with this change I get the following error:

```bash
grouped write of manifest, lock and vendor: failed to export knative.dev/pkg: fatal: failed to unpack tree object 7b4cf0bfe593195857597c822a1fbce8ce198dea
: exit status 128
```

Do we also need to update Gopkg.lock or?

houshengbo:
Just updated this PR. Could you try one more time? @greghaynes

greghaynes:
Works, thanks!

/lgtm
/approve

Issue #39, opened 2019-06-26
Discussion: Operator cli installation steps


trshafer:
There are two fundamental pieces to consider:
1. Ease of installation. From the [2019-06-24 meeting notes](https://docs.google.com/document/d/1DoCG0VGZ0_Nj84Ci443bEefSOnJAWVDXUbcYR-viVfY/edit#heading=h.g130s4kmuni6): "Plain install, it should just work" 
2. Ease of customization. Having the operator just work when installed, might lead to unexpected results if the proper upfront customization isn't handled.

### Option 1
The user must apply the CRD and then the operator checks for the CR and applies the default one if one does not automatically exist. https://github.com/knative/serving-operator/blob/6c0272eca46470454567e78aafbfd53cf7a8f53c/pkg/controller/knativeserving/knativeserving_controller.go#L298-L300

One disadvantage of this flow is there is no way to install the operator in the cluster and not have the knative installed, essentially have the operator in a wait state. We could modify the `KnativeServing` CRD to have an `enabled` flag, which supports an deactivated operator.

```
kubectl apply -f config/crds/serving_v1alpha1_knativeserving_crd.yaml
Deploy the image such as ko apply -f config/
```

**The is the current flow.**

If the user wants to apply a non-default CR, the flow is:

```
kubectl apply -f config/crds/serving_v1alpha1_knativeserving_crd.yaml
kubectl apply -f config/crds/serving_v1alpha1_knativeserving_cr.yaml // mutate this file, or apply a separate cr.
Deploy the image such as `ko apply -f config/
```

### Option 2
Another option is to have the operator install the CRD, but the user installs the CR. Because the operator now requires a second step to "activate" it, this makes it easier to have the operator live in a cluster without being activated. This is also a nice flow if the user wishes to customize the CR before applying it as there is an explicit step to now start installing.

```
Deploy the image such as `ko apply -f config/`
kubectl apply -f config/crds/serving_v1alpha1_knativeserving_cr.yaml
```

### Option 3

The operator will install everything for the user.
```
Deploy the image such as `ko apply -f config/`
```

### Option 4

The user needs to apply both the CR and the CRD.
```
kubectl apply -f config/crds/serving_v1alpha1_knativeserving_crd.yaml
Then deploy the image such as `ko apply -f config/`
kubectl apply -f config/crds/serving_v1alpha1_knativeserving_cr.yaml
```

greghaynes:
[...]

PR #55, opened 2019-07-08
Update default CR with knative-serving namespace.

trshafer:
knative-serving is the namespace used by the knativeserving controller to look
up the KnativeServing CR.

I'm not sure if there should be another change to add a namespace to knativeserving_controller: https://github.com/knative/serving-operator/blob/94e6137bb6df889c3a386435ab80a390ba66419b/pkg/controller/knativeserving/knativeserving_controller.go#L309.

### DIFF
--- a/config/crds/serving_v1alpha1_knativeserving_cr.yaml
+++ b/config/crds/serving_v1alpha1_knativeserving_cr.yaml
@@ -2,6 +2,7 @@ apiVersion: serving.knative.dev/v1alpha1
 kind: KnativeServing
 metadata:
   name: knative-serving
+  namespace: knative-serving
 spec:
   config:
     defaults:


trshafer:
Ack.

What I find confusing is executing:
`kubectl apply -f config/crds/serving_v1alpha1_knativeserving_cr.yaml` does nothing, as the namespace is not present in the CR.

jcrossley3:
Hmm. Without adding `-n knative-serving` to that command, I'd expect it to show up in `default`, which should cause the operator to install KS there, iirc.

trshafer:
https://github.com/knative/serving-operator/blob/94e6137bb6df889c3a386435ab80a390ba66419b/pkg/controller/knativeserving/knativeserving_controller.go#L140 causes the controller to bail out unless the namespace is `knative-serving`. The controller (currently) only knows how to operate on `KnativeServing` within the `knative-serving` namespace.

jcrossley3:
Ah, right, I forgot about that. So how we proceed here greatly depends on our resolution to https://github.com/knative/serving-operator/issues/39 but I personally view `isInteresting` as a bit of a desperate hack to prevent folks from having to make any decisions whatsoever. :stuck_out_tongue_winking_eye: 

I'm curious, though: since the operator is going to try and create that CR for you in the namespace you're adding in this PR, why are you manually applying it?

PR #56, opened 2019-07-08
Add the code generation of injection

houshengbo:
* This PR also adds the generation of "injection" for further reconciler.


Gopkg.toml
```diff
@@ -12,6 +12,8 @@ required = [
   "github.com/knative/test-infra/scripts",
   "github.com/knative/test-infra/tools/dep-collector",
   "sigs.k8s.io/controller-tools/pkg/crd/generator",
+  # For load testing.
+  "github.com/rakyll/hey"
```
greghaynes: Do we need to add this dep, is it being used anywhere here?

houshengbo: PLZ check the issue #57.

k4leung4:
https://github.com/knative/test-infra/issues/770

This can be manually overrided with "cla: yes"
https://github.com/knative/test-infra/issues/770#issuecomment-494506556

houshengbo:
cla: yes

k4leung4:
@adrcunha Do you know why the CLAbot is complaining about "author consent"?

adrcunha:
> @adrcunha Do you know why the CLAbot is complaining about "author consent"?

No. Commits were force pushed 6 times, then the error happened. Then there were 2 more force pushes, and the latest one seems OK. My theory is that one of the 6 force pushes had a wrong user. However, since there's no history to prove or reject my theory, it's just a theory. I'll force the CLA and let's move on.

k4leung4:
It seems like @adrcunha theory might be correct, that some of the force commits may have overridden ones with co-authors, which we might not be able to bypass.

@adrcunha to confirm.

adrcunha:
> It seems like @adrcunha theory might be correct, that some of the force commits may have overridden ones with co-authors, which we might not be able to bypass.
> @adrcunha to confirm.

As I said, I can't confirm it because there's no history. But the new label should solve the issue anyways.

k4leung4:
@googlebot rescan

adrcunha:
@googlebot I consent.

PR #65, opened 2019-07-10
Add ability to ignore a KnativeServing CR.

trshafer:
This prevents the operator from mutating the knative serving installation state.
If knative serving is uninstalled, it will remain uninstalled.
If knative serving is installed, it will remain installed.

Addresses #39

### DIFF
--- a/pkg/apis/serving/v1alpha1/knativeserving_types.go
+++ b/pkg/apis/serving/v1alpha1/knativeserving_types.go
@@ -67,6 +67,11 @@ type KnativeServingSpec struct {
 	// +optional
 	Registry Registry `json:"registry,omitempty"`
 
+	// A means to prevent installation of the knative services.
+	// By default knative serving will be installed. Setting ignore to true, will cause the operator not to install knative serving.
+	// +optional
+	Ignore bool `json:"ignore,omitempty"`
+
 	// A means to override the knative-ingress-gateway
 	KnativeIngressGateway KnativeIngressGateway `json:"knative-ingress-gateway,omitempty"`
 }
--- a/config/crds/serving_v1alpha1_knativeserving_crd.yaml
+++ b/config/crds/serving_v1alpha1_knativeserving_crd.yaml
@@ -51,6 +51,10 @@ spec:
               description: A means to override the corresponding entries in the upstream
                 configmaps
               type: object
+            ignore:
+              description: A means to prevent installation of the knative services.
+                By default knative serving will be installed. Setting ignore to true, will cause the operator not to install knative serving.
+              type: boolean
             knative-ingress-gateway:
               description: A means to override the knative-ingress-gateway
               type: object
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -157,6 +157,11 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, r.ignore(instance)
 	}
 
+	if instance.Spec.Ignore {
+		reqLogger.V(1).Info("KnativeServing is intentionally ignored")
+		return reconcile.Result{}, r.itentionallyIgnore(instance)
+	}
+
 	stages := []func(*servingv1alpha1.KnativeServing) error{
 		r.initStatus,
 		r.install,
@@ -314,6 +319,12 @@ func (r *ReconcileKnativeServing) ignore(instance *servingv1alpha1.KnativeServin
 		err = r.updateStatus(instance)
 	}
 	return
+}
+
+// Reflect our ignorance in the KnativeServing status
+func (r *ReconcileKnativeServing) itentionallyIgnore(instance *servingv1alpha1.KnativeServing) (err error) {
+	instance.Status.MarkIgnored("Intentionally ignored")
+	return r.updateStatus(instance)
 }
 
 // If we can't find knative-serving/knative-serving, create it


pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -60,6 +60,11 @@ type KnativeServingSpec struct {
 	// If no registry is provided, the knative release images will be used.
 	// +optional
 	Registry Registry `json:"registry,omitempty"`
+
+	// A means to prevent installation of the knative services.
+	// By default knative serving will be installed. Setting ignore to true, will cause the operator not to install knative serving.
+	// +optional
+	Ignore bool `json:"ignore,omitempty"`
```
houshengbo: So if we set Ignore to false, it will be equivalent to skipping this field?

trshafer: Correct. The default is false. I don't want there to be a difference between a no value and a false. I don't know if go will support a trinary of nil,true,false.

I imagine the usage will be either provide "ignore: true" or leave it out entirely.

evankanderson: If you want a trinary value, you need to do something like [corev1.ConditionStatus](https://godoc.org/k8s.io/api/core/v1#ConditionStatus), I think.

Do we want the default to be "active" or "inactive"? If we want the default to be "active", then I think `Ignore` or `Disable` are good field names. If we want the default to be "inactive", then `Enable` or `Activate` are probably good names. (You want the zero-value `false` condition to line up with the name.)

I would probably slightly prefer `Disable` over `Ignore`, I think, but I leave it to your judgement.

jcrossley3:
/hold
Hey my first hold! ^^

I'm uncomfortable with this. I'm unclear exactly what problem it's solving. I think most people think of operators as processes watching and reacting to changes in a custom resource. So to me, if I want the operator to ignore the CR, I delete the CR. Having a "ignore flag" in the CR instead seems like an abuse of the operator. What am I missing?

I think the thing that might be clouding the issue is that the operator currently creates a CR when it starts up. That is also, arguably, an abuse of the operator. I feel like giving the user control of its creation may obviate the desire for the "ignore flag". Maybe?

trshafer:
Hey @jcrossley3 WRT https://github.com/knative/serving-operator/pull/65#issuecomment-510235492, I think either we need some sort of an ignore flag or the operator should not by default install the CR (see https://github.com/knative/serving-operator/issues/39#issuecomment-506990336). There's no way for the user to have the operator in a wait state right now. (1) 

I'm not sold on the idea that the "ignore" flag should still keep around the existing knative services if it has already been installed.

(1) Well the user could install the operator, which would then install knative-serving and then the user could delete the CR after it has been created which would delete knative-serving. But I think that's a weird usecase.

pkg/apis/serving/v1alpha1/knativeserving_types.go
trshafer: The only time I've found a trinary useful is when you really need to distinguish between "unknown" and "false". I don't think in this instance we want to distinguish between unknown and false, because according to the the discussion in https://github.com/knative/serving-operator/issues/39, the general consensus is that installing the operator should install knative by default. I think we want "unknown" to be the same as "install knative serving". Deploy the CRD and the operator and then knative is installed.

jcrossley3:
I think we're talking ourselves into changing the operator so that it does _not_ auto-create the CR. I'm pretty sure @bbrowning is the only person who wants that, so let's see if he _still_ wants it after considering an "ignore flag" in the CR. :stuck_out_tongue_winking_eye:

trshafer:
Summoning @bbrowning!

vincent-pli:
@trshafer @jcrossley3 
I also need a flag ^ ^, but not flag in `cr`.
i think if create `cr` automatic is the behavior of `operator` itself, should be defined in `operator` rather that `cr`
so
I prefer to add flag in `operator.yam`, like this:
```
    spec:
      serviceAccountName: knative-serving-operator
      containers:
        - name: knative-serving-operator
          image: github.com/knative/serving-operator/cmd/manager
          imagePullPolicy: IfNotPresent
          env:
            - name: AUTOMITIC_INSTALL
              value: false
```

BTW, I don't like the idea to let user create a customized `cr` before `operator` installation to stop install a `build-in` `cr`

jcrossley3:
> i think if create `cr` automatic is the behavior of `operator` itself, should be defined in `operator` rather that `cr` so I prefer to add flag in `operator.yaml`

In an early version of the operator, we had an `--auto-create` flag. It was boolean for a while (defaulted to false), and then it took a namespace as an argument (where to put the CR), and finally it was removed (rather than defaulting it to true) just to simplify the release process.

Any particular reason to favor an env var over an arg to signal the operator to ensure a CR exists? And would that be enough to put the kibosh on this PR?

vincent-pli:
@jcrossley3 like this?
```
	flag.BoolVar(&autoInstall,
		"auto-install",
		false,
		"Automatically create an install custom resource (install pipeline)",
	)
```
arg or env parameter are all ok for me, just need the feature,  could we add this back?

trshafer:
I don't really like the idea of changing an argument or environment variable to the deployment. This would require re-deploying the operator and changing the operator yaml, while changing a CR is much easier of a task. Asking the user to mutate the operator yaml I think is more surprising. Most people just install things like `kubectl apply -f some/file.yaml` or `kubectl apply -f https://some-remote/file.yaml`. Asking users to download the yaml and change the Deployment is more confusing IMO than having an example CR that they can mutate and apply.

jcrossley3:
I agree changing an example CR rather than a Deployment is user-friendlier. If we expect most users to need to do this, then I don't think the operator should auto-create the CR.

bbrowning:
The automatic installation of the CR was a convenience to get the install as close to a single step as possible. If there are valid reasons to not want the default installation then I'd prefer we just stop the automatic installation instead of adding an ignored flag.

My overall desire was the get the installation of Knative Serving as simple as possible. If we can still provide a simple install experience without automatic installation, then I won't object to that.

vincent-pli:
@trshafer @jcrossley3 
Could I add `auto-install` back?
I have no idea why it removed, but seems it's make sense to handle some `D2` installation case.

trshafer:
@vincent-pli, is `auto-install` a command argument to the deployment?

vincent-pli:
@trshafer 
yes, command argument of deployment of `operator`

trshafer:
@bbrowning the rationale not to force an installation is so someone can install the operator on a cluster without forcing the installation of knative. The reason this is desirable is so a cluster can be given to someone that is setup to install knative but doesn't have knative running. A minimal switch of either installing the `KnativeServing` (assuming no auto install) CR or mutating the `ignored` flag is easier than applying a new operator.

I'm somewhat talking myself out of this feature and thinking installation of the operator via applying yaml is enough of a flow.

jcrossley3:
/hold
How about we close this PR and then restore the `--auto-install` flag with a default value of false? I expect very few people will ever set it to true (possibly only @vincent-pli :smile:) and we instruct users to 1) install the operator and then b) create the CR. We could even provide a yaml that contains both and more sophisticated users could break them up.

trshafer:
Closing in-favor of --auto-install

PR #69, opened 2019-07-15
Rename the directory controller into reconciler

houshengbo:
This PR is the the first step of the refactoring to use knative/pkg.


houshengbo:
@k4leung4 @trshafer @greghaynes @jcrossley3 
Let's me do the renaming first. I plan to add the test coverage after the refactoring is done, since we will have to change the test cases anyhow, even if we create the test cases in the midst of refactoring.

PR #70, opened 2019-07-15
Update the branch into client-go for the package jcrossley3/manifestival

houshengbo:
The purpose of this PR is to get rid of the usage of client.Client in controller-runtime.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/restmapper"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
+	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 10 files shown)


cmd/manager/main.go
```diff
@@ -19,14 +19,15 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"k8s.io/client-go/tools/clientcmd"
```
k4leung4: nit: group with other k8s import

houshengbo: done

greghaynes:
I really dont think this should pull in 85k of vendor updates? Is it possible you ran update-deps while this was checked out in the old (github.com/knative/serving-operator) path?

greghaynes:
@jcrossley3 Do you want to keep this as a separate branch in manifestival repo or are you interested in making this master or?

houshengbo:
@greghaynes When we change the branch from master to client-go for github.com/jcrossley3/manifestival, all the changes happened accordingly, if we run ./hack/update-deps.sh.

PR #74, opened 2019-07-16
openshift: Add env value to disable openshift ingress extension

nak3:
When serving operator is installed on OpenShift env, it always
installs openshift-ingress as well. This is sometimes painful when we
try to use knative's default ingress or when we would like to test
other version of openshift-ingress.

To address such pains, this patch introduces
`DISABLE_OPENSHIFT_INGRESS` env to disable installation of
openshift-ingress.


nak3:
Closing this PR.

Issue #21, opened 2019-06-21
Validation Script


trshafer:
## In what area(s)?

> /area API
> /area autoscale
> /area build
> /area monitoring
> /area networking
/area test-and-release

## Describe the feature
It would be helpful to have a validation script that took in a kubeconfig and gave a pass/fail or likelihood of passing if knative could be run on the cluster. The script would be run by a cluster admin and validate that the kubernetes cluster is properly setup to support knative. 

It would be run like: `validate-cluster.sh --kubeconfig=/path/to/kubeconfig`

I'm not immediately sure what all the checks would be, one initial suggestion by @dgerd is the initial check would simply check the k8s version and perhaps a version of a supported ingress/gateways services.

yu2003w:
I would like to work on this item.

houshengbo:
@yu2003w Are u a member of Knative? Why cannot I assign it to you?

yu2003w:
@houshengbo I'm not member of knative.

houshengbo:
@yu2003w You definitely need to join the knative community as a member. However, you can start working on it anyhow

@k4leung4 @greghaynes Since we folks have gone the process of proposing knative member before, could you propose @yu2003w as a member of knative? I can be another member proposing @yu2003w if needed.

yu2003w:
validation scripts check following items:
1,  k8s cluster version
2, Istio version and following services deployed
    cluster-local-gateway
    istio-ingressgateway
3, cert manager installed

greghaynes:
Thanks for working on this. What do we think about rather than a shell script we make this go code which our operator can run? I expect we will want our operator to perform checks like this as part of normal operation and it will be difficult to bake in a shell script which it runs.

yu2003w:
Operator should perform the checks before installing Knative Serving.
Does operator need to check the conditions such as ingress gateway periodically after knative serving installed?
Does operator support to install serving with Gloo preinstalled rather than istio?

trshafer:
> Operator should perform the checks before installing Knative Serving.
Yes. I am open to the idea of updating the `KnativeServing` CRD to add the desired dependency configuration. @houshengbo  @greghaynes and @trshafer jcrossley3 

Something like this?

``` yaml
spec:
  deps:
    istio: true
    gloo: false
    mtls: false
```

> Does operator need to check the conditions such as ingress gateway periodically after knative serving installed?
[...]

PR #75, opened 2019-07-16
Validation before installing knative serving

yu2003w:
To fix problem #21 

Signed-off-by: Wei Yu(Jared) <yuweiw@cn.ibm.com>

### DIFF
--- a/config/crds/serving_v1alpha1_knativeserving_crd.yaml
+++ b/config/crds/serving_v1alpha1_knativeserving_crd.yaml
@@ -93,6 +93,13 @@ spec:
                       name:
                         description: The name of the secret.
                         type: string
+            knative-dependency:
+              description: Dependencies for installing knative serving, operator need to check whether
+                dependencies are met before invoking install knative serving.
+              type: array
+              items:
+                type: string
+                description: Dependent component name
           type: object
         status:
           description: Status defines the observed state of KnativeServing
--- a/pkg/reconciler/knativeserving/common/validation.go
+++ b/pkg/reconciler/knativeserving/common/validation.go
@@ -1,0 +1,174 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package common
+
+import (
+	"fmt"
+	"strings"
+
+	"go.uber.org/zap"
+
+	"github.com/rogpeppe/go-internal/semver"
+	v1beta1 "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/pkg/version"
+	opversion "knative.dev/serving-operator/version"
+)
+
+const (
+	// IstioDependency name
+	istioDependency     = "istio"
+	istioNamespace      = "istio-system"
+	istioPilotContainer = "istio-proxy"
+	// CertManagerDependency name
+	certManagerDependency = "certmanager"
+	certManagerNamespace  = "cert-manager"
+)
+
+type depMeta struct {
+	// dependent
+	name string
+	// in which dependent installed
+	namespace string
+	// deployments installed in the namespace
+	deploy []string
+}
+
+var depMetas = map[string]depMeta{
+	istioDependency: {
+		name:      istioDependency,
+		namespace: istioNamespace,
+		deploy: []string{
+			"istio-ingressgateway",
+			"cluster-local-gateway",
+		},
+	},
+	certManagerDependency: {
+		name:      certManagerDependency,
+		namespace: certManagerNamespace,
+		deploy: []string{
+			"cert-manager",
+		},
+	},
+}
+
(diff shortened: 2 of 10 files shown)


config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -74,6 +74,17 @@ spec:
                   type: object
                   additionalProperties:
                     type: string
+              knative-dependency:
+                description: Dependencies for installing knative serving, operator need to check whether
```
mattmoor-sockpuppet:
```suggestion
                description: Dependencies for installing knative serving, operator need to check whether
```

mattmoor-sockpuppet:
```suggestion
                description: Dependencies for installing knative serving, operator need to check whether
```

config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -74,6 +74,17 @@ spec:
                   type: object
                   additionalProperties:
                     type: string
+            knative-dependency:
+              description: Dependencies for installing knative serving, operator need to check whether
```
mattmoor-sockpuppet:
```suggestion
              description: Dependencies for installing knative serving, operator need to check whether
```

mattmoor-sockpuppet:
```suggestion
              description: Dependencies for installing knative serving, operator need to check whether
```

jcrossley3:
/hold
I don't think we want dependency versions in the the CR spec. We should already have a mapping of KS release to supported versions of dependencies. Exposing that to the user is just noise, I think.

yu2003w:
@jcrossley3  I agree with you that we should already have the mapping.
But How could serving-operator get the mapping information? Configure that in CM or hard-code for verifying dependency version? 

also CC'ed @trshafer Do you have any advice?

trshafer:
If we're hardcoding the version of knative serving that gets installed, I think it's ok for us to hard code the supported dependencies. It would be nice if these versions were easy to locate within the repo as I imagine people would be interested in looking at what the supported dependency versions were.

jcrossley3:
I would think easy access to dependency version mappings is a question for knative-serving itself to answer, no? The operator would consume that information just like a human would.

yu2003w:
@jcrossley3 Yes, kn serving should know dependencies versions information. I would like to raise a question in kn serving chat group.
Another question is that if I override registry image with kn serving 0.6.1 when creating CR, it could installed successfully. And it seems that we don't have a mechanism to verify that. Is it necessary for serving-operator to validate this?

jcrossley3:
> @jcrossley3 Another question is that if I override registry image with kn serving 0.6.1 when creating CR, it could installed successfully. And it seems that we don't have a mechanism to verify that. Is it necessary for serving-operator to validate this?

Good question. My initial thought is that once you change that image, all bets are off, and you're responsible. It makes me wonder whether we should just disable all dependency checks when an image is overridden.

pkg/reconciler/knativeserving/common/validation.go
```diff
@@ -0,0 +1,276 @@
...
+	basemsg := t.meta.name + " validation"
+	// check namespace created
+	_, err := client.CoreV1().Namespaces().Get(t.meta.namespace, metav1.GetOptions{})
+	if err != nil {
+		msg := t.meta.namespace + " doesn't exist"
+		log.Error(err, basemsg, "namespace checking", msg)
+		return fmt.Errorf(msg)
+	}
+
+	// for istio, version could be get frome env of container istio-proxy in pod istio-pilot
```
mattmoor-sockpuppet:
```suggestion
	// for istio, version could be get from env of container istio-proxy in pod istio-pilot
```

yu2003w: > I would think easy access to dependency version mappings is a question for knative-serving itself to answer, no? The operator would consume that information just like a human would.

@jcrossley3 I didn't find information about how to read kn serving mapping with dependency version by operator. Do you have any advice? Or did I misunderstand your comments?

yu2003w:
@jcrossley3 I didn't find information about how to read kn serving mapping with dependency version by operator. Do you have any advice? Or did I misunderstand your comments?

> I would think easy access to dependency version mappings is a question for knative-serving itself to answer, no? The operator would consume that information just like a human would.

pkg/reconciler/knativeserving/common/validation.go
```diff
@@ -0,0 +1,260 @@
...
+	"k8s.io/client-go/tools/remotecommand"
+	"knative.dev/pkg/version"
+	opversion "knative.dev/serving-operator/version"
+)
+
+const (
+	// IstioDependency name
+	istioDependency     = "istio"
+	istioNamespace      = "istio-system"
+	istioProductVersion = "productVersion"
```
markusthoemmes: This seems unused.

pkg/reconciler/knativeserving/common/validation.go
```diff
@@ -0,0 +1,260 @@
...
+	namespace string
+	// deployments installed in the namespace
+	deploy []string
+}
+
+var depMetas = map[string]depMeta{
+	istioDependency: {
+		name:      istioDependency,
+		namespace: istioNamespace,
+		deploy: []string{
```
markusthoemmes: Why do we need this list? we don't really depend on an istio installation even if we depend on istio.

markusthoemmes (commented):
I think this should be approached in a more abstract way, at least for Istio. We don't really need to depend on specific deployments being there etc. Really all we depend on at this point is:

1. Some CRDs in a version that we need.
2. The configured gateways (istio-ingressgateway and cluster-local-gateway) installed and at the location defined in Knative's config.

pkg/reconciler/knativeserving/common/validation.go
yu2003w: We don't need to know the details of Istio deployments. I will update PR soon.

houshengbo:

Issue #4, opened 2019-06-11
Operator should use knative/pkg controller


greghaynes:
Currently the operator uses controller-runtime / operator-sdk. We should retool it on top of Knative's controller lib to more closely resemble the other Knative operators.

houshengbo:
We need to first generate the client code for the knativeserving resource under pkg/api.
Then, rewrite both of the controller package and main.go about how to create & start the controller.

houshengbo:
We have already resolved it, by finishing the structural transition from controller-runtime to knative/pkg.

PR #77, opened 2019-07-16
Change the branch from master to client-go for dependency jcrossley3/manifestival 🧗‍♂️

houshengbo:
Co-authored-by: Markus Thömmes \<markusthoemmes@me.com\>
## Proposed Changes

* This PR replaces the clients of controller-runtime with either KS client or kubeclient.
* This PR regenerates the dependencies.

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/dynamic"
@@ -108,7 +109,7 @@ type ReconcileKnativeServing struct {
 // Create manifestival resources and KnativeServing, if necessary
 func (r *ReconcileKnativeServing) InjectClient(c client.Client) error {
 	koDataDir := os.Getenv("KO_DATA_PATH")
-	m, err := mf.NewManifest(filepath.Join(koDataDir, "knative-serving/"), *recursive, c)
+	m, err := mf.NewManifest(filepath.Join(koDataDir, "knative-serving/"), *recursive, r.clientConfig)
 	if err != nil {
 		log.Error(err, "Failed to load manifest")
 		return err
@@ -135,7 +136,7 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 	instance := &servingv1alpha1.KnativeServing{}
 	if err := r.client.Get(context.TODO(), request.NamespacedName, instance); err != nil {
 		if errors.IsNotFound(err) {
-			r.config.DeleteAll()
+			r.config.DeleteAll(&metav1.DeleteOptions{})
 			reqLogger.V(1).Info("No KnativeServing")
 			return reconcile.Result{}, nil
 		}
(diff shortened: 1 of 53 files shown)


houshengbo:
@bbrowning 
We used to create the manifestval with 
```
mf.NewManifest(filepath.Join(koDataDir, "knative-serving/"), *recursive, c)
```, but now we switch to use

```
mf.NewManifest(filepath.Join(koDataDir, "knative-serving/"), *recursive, r.clientConfig)
```

The last(third) param we used before was the client.Client in controller-runtime. I guess it did some verification or retry to make sure the manifestival loads the updated information about the CRDs in kube. However, the param r.clientConfig is just a config, which cannot make sure the manifestival has the up-to-date info regarding the CRDs in kube. This is my guess about the reason it fails.

**Comment by bbrowning** [comment]
I was able to reproduce this locally on minikube and can confirm that the `Image` CRD is defined but the controller reconcile loop can never find it. However, if I delete the serving operator pod, then the reconcile loop succeeds and successfully installs Knative Serving.

This adds weight that the client is caching some list of installed CRDs in the cluster and we need to either wait for CRDs to be installed before creating instances of those CRs or refresh the client's cache when we hit this error. I'll try to find out how to do one of those two.

File: pkg/reconciler/knativeserving/knativeserving_controller.go [hunk 1]
```diff
@@ -62,7 +64,8 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileKnativeServing{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	clientConfig, _ := clientcmd.BuildConfigFromFlags("", "")
```
trshafer: What are these flags for?

houshengbo: Was trying to use a config passed from the manager, but created it here for a quick test.
should be masterurl and kubeconfig passed into.

bbrowning:
I was able to confirm that this is a client caching issue and workaround it by the hacky solution below that indirectly calls `NewManifest` on manifestival that creates new client-go Clients. This isn't a solution I'd propose as the right one, but it does get things working.

We need some way to tell the client to refresh its discovery mappings after the Knative Serving CRDs are installed in the cluster.

```diff
diff --git a/pkg/reconciler/knativeserving/knativeserving_controller.go b/pkg/reconciler/knativeserving/knativeserving_controller.go
index 1092ba0..f69d902 100644
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -205,6 +206,9 @@ func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServi
                err = extensions.PreInstall(instance)
                if err == nil {
                        err = r.config.ApplyAll()
+                       if err != nil {
+                               r.InjectClient(r.client)
+                       }
                        if err == nil {
                                err = extensions.PostInstall(instance)
                        }
```

houshengbo:
Let me tentatively put this quick fix here.
After https://github.com/knative/serving-operator/issues/4 is implemented, we probably do not need this fix any more.

pkg/reconciler/knativeserving/knativeserving_controller.go
houshengbo: Check https://github.com/knative/serving-operator/pull/78 for a better way to access the cfg. May use that one instead.

jcrossley3:
FWIW, we had to fix operator-sdk (actually controller-runtime) similarly: https://github.com/operator-framework/operator-sdk/issues/1328

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -201,6 +205,9 @@ func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServi
 		err = extensions.PreInstall(instance)
 		if err == nil {
 			err = r.config.ApplyAll()
+			if err != nil {
```
greghaynes: Can we add a link / a description of why were doing this (to force our client to refresh its CRD discovery mappings)?

houshengbo: Done.

markusthoemmes: The root cause of this seems to be that Manifestival's RESTMapper doesn't refresh itself. The operator-sdk provides such a mapper.

I think https://github.com/jcrossley3/manifestival/blob/2d595bb74f312fb5f9228ade0de5dbf792840061/manifestival.go#L54 needs to change to said mapper and we might not need this workaround.

jcrossley3: Doh! Why didn't I think of that?!?! Good call. I'll take care of it!

jcrossley3: I have updated the `client-go` branch of manifestival to use a dynamic RESTMapper that will obviate the need for this workaround. There should be no need to refresh the client anymore.

jcrossley3:
/hold
This change is obsolete in light of #94 after the `client-go` manifestival dep is updated.

houshengbo:
@jcrossley3 @markusthoemmes Shall we unhold this one to go in?

jcrossley3:
> @jcrossley3 @markusthoemmes Shall we unhold this one to go in?

Check the deps. I get errors from `hack/update-deps.sh`

jcrossley3:
Nevermind. My `vendor/` was screwed up.
/hold cancel

houshengbo:
@jcrossley3 Based on the discussion with @markusthoemmes in the past: when vendor screwed up, run the following commands:
```
rm -rf vendor/
rm Gopkg.lock
dep ensure -v
./hack/update-deps.sh
```

Shall we create a script named regenerate-deps.sh to all the steps?

PR #78, opened 2019-07-16
Access the resources with kubeclient and dynamicclient

houshengbo:
Closes: #78


greghaynes:
this LGTM - I dont really have insight to the openshift bits or a way to test them though

PR #83, opened 2019-07-18
Reimplement pkg/controller as the backbone of operator

houshengbo:
* This PR refactors the backbone of the operator, by replacing the manager with pkg/controller.
* Reconciler is implemented according for pkg/controller as well.
* By far, we still keeps the creation of the manager for purpose of getting the client of controller-runtime, since this client has not yet completely been replaced.

Closes: #4


config/config-logging.yaml
```diff
@@ -0,0 +1,70 @@
+# Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
# Copyright 2019 The Knative Authors
```

config/config-observability.yaml
```diff
@@ -0,0 +1,100 @@
+# Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
# Copyright 2019 The Knative Authors
```

config/operator.yaml
```diff
@@ -26,3 +30,16 @@ spec:
...
+              value: knative.dev/serving-operator
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: config-observability
+          volumeMounts:
+            - name: config-logging
+              mountPath: /etc/config-logging
+      volumes:
+        - name: config-logging
+          configMap:
+            name: config-logging
```
mattmoor-sockpuppet:
```suggestion
            name: config-logging

```

config/config-logging.yaml
mattmoor-sockpuppet:
```suggestion
# Copyright 2019 The Knative Authors
```

config/config-observability.yaml
mattmoor-sockpuppet:
```suggestion
# Copyright 2019 The Knative Authors
```

config/operator.yaml
mattmoor-sockpuppet:
```suggestion
            name: config-logging

```

config/config-logging.yaml
mattmoor-sockpuppet:
```suggestion
# Copyright 2019 The Knative Authors
```

config/config-observability.yaml
mattmoor-sockpuppet:
```suggestion
# Copyright 2019 The Knative Authors
```

config/operator.yaml
mattmoor-sockpuppet:
```suggestion
            name: config-logging

```

jcrossley3:
/hold
In testing this locally, I noticed that the operator doesn't
recreate a deployment (say, activator) if I manually delete it. And
it's difficult for me to diagnose since logging seems to be broken.

jcrossley3:
Something else I noticed: the status of the `KnativeServing` CR never becomes "Ready", as it never recognizes that its deployments are ready. This may be related to why deleted deployments aren't recreated. I confirmed that the owner references are there, but it doesn't seem like events from the children are received by the parent.

houshengbo:
@jcrossley3 I restore all the code in knativeserving_controller.go. I expect all the KnativeServing CRs should be ready at least.

I am not sure if ./hack/run-local.sh works only for operator-sdk?

jcrossley3:
> I am not sure if ./hack/run-local.sh works only for operator-sdk?

It does currently, but it should be possible to change it to work with pkg/controller. @markusthoemmes has assured me it's possible to run it as a remote client to a cluster.

markusthoemmes:
To be fair, the shared components might need some love to make that possible. The config loading might need to be a bit more elaborate than today.

houshengbo:
@jcrossley3 @markusthoemmes Let's first make sure this PR works fine with `ko`.

Regarding run-local.sh, we can work with a separate issue, since shared component may enable the support as a base.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -77,11 +78,11 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 	logger.Info("Let us do a reconcile.")
 
-	// Get the Route resource with this namespace/name.
+	// Get the KnativeServing resource with this namespace/name.
 	original, err := c.knativeServingLister.KnativeServings(namespace).Get(name)
 	if apierrs.IsNotFound(err) {
 		// The resource may no longer exist, in which case we stop processing.
-		logger.Errorf("route %q in work queue no longer exists", key)
+		logger.Errorf("KnativeServing %q in work queue no longer exists", key)
 		return nil
```
jcrossley3: This is the point at which you should invoke `DeleteAll` on the manifest to clean up the installed resources, I think.

houshengbo: Is this used to resolved the installation of all the knative resources?

houshengbo:
@jcrossley3 The commit "Rewrite the checkDeployment function" resolves the issue of the deployment recreation. I am adding a test case to cover it now.

pkg/reconciler/knativeserving/knativeserving.go
jcrossley3: I'm not sure I understand your question, but I assume `Reconcile` gets called when some lifecycle event occurs for the KnativeServing instance (or one of its children), so if it's not found, logically that means it's been deleted, correct? At which point, we should clean up its resources.

PR #84, opened 2019-07-18
Add ability to skip certmanager and networking-istio

### LINKED ISSUE #82: Disable installation of certain deployments
Some installations do not want the certain deployments, such as networking-certmanager or networking-istio. 

This can be accomplished by adding additional configuration to the `KnativeServing` CRD.

## Proposal 1
Proposal 1 supports only skipping deployments. This would be look for the `networking-istio` name field of the deployment.

```yaml
spec:
  deployments:
    networking-istio: false
```

## Proposal 2

Another proposal is to add the ability to skip any resource:

```yaml
spec:
  skip:
    - resource: apps/v1/Deployment 
      name: networking-istio
```

trshafer:
Fixes #82

### DIFF
--- a/config/crds/serving_v1alpha1_knativeserving_crd.yaml
+++ b/config/crds/serving_v1alpha1_knativeserving_crd.yaml
@@ -83,6 +83,16 @@ spec:
                   type: object
                   additionalProperties:
                     type: string
+            disabled-components:
+              description: A means to skip specific sections in the installation
+              type: object
+              properties:
+                certmanager:
+                  type: boolean
+                  description: The ability to disable the certificate deployment and related assets.
+                networking-istio:
+                  type: boolean
+                  description: The ability to disable the networking-istio deployment.
           type: object
         status:
           description: Status defines the observed state of KnativeServing
--- a/pkg/apis/serving/v1alpha1/knativeserving_types.go
+++ b/pkg/apis/serving/v1alpha1/knativeserving_types.go
@@ -51,6 +51,15 @@ type IstioGatewayOverride struct {
 	Selector map[string]string `json:"selector,omitempty"`
 }
 
+// DisabledComponents skips certain sections of knative serving
+type DisabledComponents struct {
+	// The ability to disable the certificate deployment and related assets.
+	Certmanager bool `json:"certmanager,omitempty"`
+
+	// The ability to disable the networking-istio deployment.
+	NetworkingIstio bool `json:"networking-istio,omitempty"`
+}
+
 // KnativeServingSpec defines the desired state of KnativeServing
 // +k8s:openapi-gen=true
 type KnativeServingSpec struct {
@@ -72,6 +81,9 @@ type KnativeServingSpec struct {
 
 	// A means to override the cluster-local-gateway
 	ClusterLocalGateway IstioGatewayOverride `json:"cluster-local-gateway,omitempty"`
+
+	// A means to skip specific sections in the installation
+	DisabledComponents DisabledComponents `json:"disabled-components,omitempty"`
 }
 
 // KnativeServingStatus defines the observed state of KnativeServing
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -221,12 +221,70 @@ func (r *ReconcileKnativeServing) executeInstall(extensions common.Extensions, i
 		return err
 	}
 
-	err = r.config.ApplyAll()
+	err = r.applyEnabledResources(instance)
 	if err != nil {
 		return err
 	}
 
 	return extensions.PostInstall(instance)
+}
+
+func (r *ReconcileKnativeServing) applyEnabledResources(instance *servingv1alpha1.KnativeServing) error {
+	for _, spec := range r.config.Resources {
+		if err := r.applyOrDeleteSpec(spec, instance); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+func (r *ReconcileKnativeServing) applyOrDeleteSpec(spec unstructured.Unstructured, instance *servingv1alpha1.KnativeServing) error {
+	if !shouldApplySpec(spec, instance) {
+		return r.deleteSpec(spec)
+	}
+	return r.applySpec(spec)
+}
+
+func (r *ReconcileKnativeServing) deleteSpec(spec unstructured.Unstructured) error {
+	if err := r.config.Delete(&spec); err != nil {
+		log.Error(err, "Error deleting spec", "spec", spec)
+		return err
+	}
+	return nil
+}
+
+func (r *ReconcileKnativeServing) applySpec(spec unstructured.Unstructured) error {
+	if err := r.config.Apply(&spec); err != nil {
+		log.Error(err, "Error applying spec", "spec", spec)
+		return err
+	}
+	return nil
+}
+
+func shouldApplySpec(spec unstructured.Unstructured, instance *servingv1alpha1.KnativeServing) bool {
+	if instance.Spec.DisabledComponents.NetworkingIstio && isNetworkingIstioComponent(spec) {
+		return false
+	}
+	if instance.Spec.DisabledComponents.Certmanager && isCertmanagerComponent(spec) {
+		return false
+	}
+	return true
+}
+
+func isCertmanagerComponent(spec unstructured.Unstructured) bool {
+	return hasLabel(spec, "networking.knative.dev/certificate-provider", "cert-manager")
+}
+
+func isNetworkingIstioComponent(spec unstructured.Unstructured) bool {
+	return hasLabel(spec, "networking.knative.dev/ingress-provider", "istio")
+}
+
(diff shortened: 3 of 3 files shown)


markusthoemmes:
I wonder why we make this so "generic". The istio or certmanager deployment can consist of more than just the deployment resource. ServiceAccounts, Roles, Configs etc come to mind.

Should we instead "model" the "certificate-provider" and "ingress-provider" and filter by the respective labelss (as we do in the kubectl apply workflow)? We could allow a special value "none" to be set, which causes the entire thing to be skipped.

trshafer:
Updated the spec to support a more abstract `components`. Only certmanager and networking-istio are currently supported.

```yaml
kind: KnativeServing
spec:
  disabled-components:
    certmanager: true
    networking-istio: true
```

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -233,6 +225,63 @@ func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServi
...
+func isCertmanagerComponent(spec unstructured.Unstructured) bool {
+	for labelName, labelValue := range spec.GetLabels() {
+		if labelName == "networking.knative.dev/certificate-provider" && labelValue == "cert-manager" {
+			return true
+		}
+	}
+	return false
+}
+
+func isNetworkingIstioComponent(spec unstructured.Unstructured) bool {
```
anniefu: What's the reason for the different detection pattern between this and isCertmanagerComponent?

Is skiping the Deployment named "networking-certmanager" not equivalent to disabling certmanager?

config/crds/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -74,6 +74,16 @@ spec:
                   type: object
                   additionalProperties:
                     type: string
+            disabled-components:
```
anniefu: Maybe skip-install is a better name? disabled-components could imply disabling some serving-operator component vs knative-serving component.

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -233,6 +225,63 @@ func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServi
...
+		return err
+	}
+	err = r.applyAllResources(instance)
+	if err != nil {
+		return err
+	}
+	return extensions.PostInstall(instance)
+}
+
+func (r *ReconcileKnativeServing) applyAllResources(instance *servingv1alpha1.KnativeServing) error {
```
anniefu: [nit] applyEnabledResources or applyIncludedResources?

Otherwise r.applyAllResources and  config.ApplyAll is confusing to differentiate.

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -267,6 +316,16 @@ func (r *ReconcileKnativeServing) checkDeployments(instance *servingv1alpha1.Kna
 	return nil
 }
 
+func deploymentSkipped(deploymentName string, instance *servingv1alpha1.KnativeServing) bool {
```
anniefu: Not used?

anniefu:
The more generic ability to skip any deployment (or any label) seems like it will be better for the long term maintenance of the serving-operator. 

Forcing the knative-serving operator code to be conscious of how configs are being defined in knative-serving seems like it will lead to code breaks and extensibility issues in the future when knative-serving components change.

pkg/reconciler/knativeserving/knativeserving_controller.go
trshafer: Thanks. Removed.

pkg/reconciler/knativeserving/knativeserving_controller.go
trshafer: Done.

trshafer:
Thanks for the review @anniefu! 

WRT the operator having a deep understanding of the knative serving yaml, I think is actually a feature, the user doesn't need to know about the underlying yaml. There are downsides to having the deployments only configuration:
1. the user must know which deployments are in knative-serving, so they must know what the underlying knative serving yaml looks like.
2. There are leftover resources which are created such as roles, service accounts, etc.

See thread: https://knative.slack.com/archives/CEPGNPHP1/p1564505791094900

config/crds/serving_v1alpha1_knativeserving_crd.yaml
trshafer: I'm open to `skip-install`, that works for me.

pkg/reconciler/knativeserving/knativeserving_controller.go
jcrossley3: I'm curious about this, too. It'd be nice if we could omit all the irrelevant resources by label, if possible, e.g. `ClusterRole` and `ConfigMap`, but I'm not sure about the `Gateway` resources that also have this label.

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -233,6 +225,63 @@ func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServi
...
+	if err != nil {
+		return err
+	}
+	return extensions.PostInstall(instance)
+}
+
+func (r *ReconcileKnativeServing) applyEnabledResources(instance *servingv1alpha1.KnativeServing) error {
+	for _, spec := range r.config.Resources {
+		if !shouldApplySpec(spec, instance) {
+			continue
```
jcrossley3: I think instead of `continue` we should put `r.config.Delete(&spec)`. I believe that will account for updates to the CR that add/remove disabled/skipped components.

trshafer: Thanks, good suggestion

trshafer:
Hey @anniefu, I realized that my previous comment sounded definitive. I wanted to provide a reference to previous discussion and points that were brought up about this topic. I'm happy to continue the conversation about not having the serving-operator have deeper coupling to the serving yaml.

anniefu:
> Hey @anniefu, I realized that my previous comment sounded definitive. I wanted to provide a reference to previous discussion and points that were brought up about this topic. I'm happy to continue the conversation about not having the serving-operator have deeper coupling to the serving yaml.

No worries! I didn't interpret your comment that way at all. I agree that the serving-operator should be as user-friendly as possible, and I'm not against keeping this PR as-is for the sake of progress.

Underneath the hood it would be nice to be more generic and skip by labels though. On the spec we would only expose specific labels as properties. 

Alternatively, we could use a generic map similar to the "override" field. Where the map would be "<label-name>: true", and in the description we provide the recommended labels to use. If people are aware of other labels outside the well-defined components, they can still pass them in to skip other components. Keys that are not valid labels would just be ignored.

trshafer:
I agree that using a label system under the hood would be beneficial. Do you think that mutating the knative-serving released yaml before creating this operator API is a blocker?

anniefu:
No, I don't think changing anything in knative-serving should block this. Making the spec use "cert-manager" instead of "certmanager" so the spec matches knative-serving labels would be good.

The only other open question I have is why when skipping the networking-istio install we're looking for the the deployment named "networking-istio" instead of resources matching the "networking-istio" label.

jcrossley3:
> I agree that using a label system under the hood would be beneficial. 

I'm +1 on a label system, and I think I might even be ok exposing the actual label names in the CR spec. That seems like the best of both worlds: a generic solution that can be clear to the user by choosing good label names.

> Do you think that mutating the knative-serving released yaml before creating this operator API is a blocker?

Are you asking whether it's ok to munge the yaml beneath `cmd/manager/kodata/knative-serving`? I'm ok with that as long as whatever we change shows up in the next release. I'd hate to get into a cycle of having to post-process every release yaml.

trshafer:
I don't think I can guarantee that changes in `cmd/manager/kodata/knative-serving` will appear in the next knative release. I can ping the serving group and get their thoughts.

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -51,6 +51,15 @@ type IstioGatewayOverride struct {
 	Selector map[string]string `json:"selector,omitempty"`
 }
 
+// DisabledComponents skips certain sections of knative serving
+type DisabledComponents struct {
```
pmorie: I'm not against this being expressable in the operator's API at all, but I'm not sure this is the best API construction to use:

- The API conventions [advise against](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) using booleans as a type for API fields
- It feels a little weird to have `spec.disabledComponents.certmanager = true` in the API in terms of polarity. I'd rather indicate disablement through a "false" sounding value.

On an entirely different note, unless we make some other change, I do not believe there is a way to "disable" use of cert-manager, unless you disable the use of admission webhooks, since cert-manager is used to mint the serving certificate from the webhook controller.

aliok:
@k4leung4 https://github.com/knative/serving-operator/issues/82 is closed, so I think this PR should be closed too?

PR #85, opened 2019-07-19
Point release documentation to knative/test-infra

adrcunha:
Fixes knative/test-infra#491


hack/release.sh
```diff
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Documentation about this script and how to use it can be found
+# at https://github.com/knative/test-infra/tree/master/ci
```
houshengbo: @adrcunha I would prefer to change the link into https://github.com/knative/test-infra/blob/master/ci/manual_release.md for release instruction especially.

adrcunha: I prefer pointing to a more generic location than getting this outdated when things change in test-infra. Furthermore, manual releases are exceptions and normally should not be used.

houshengbo: @adrcunha  Could you do a rebase removing the conflict?

adrcunha: Done, thanks for the heads up.

Issue #87, opened 2019-07-24
operator restarted due to connection to API Server not established


yu2003w:
/kind bug

knative-serving-operator pod always need to restart because connection is not established
when try to become leader. The detailed log as below,

```
{"level":"info","ts":1563951785.0586658,"logger":"cmd","msg":"Go Version: go1.12.5"}
{"level":"info","ts":1563951785.058726,"logger":"cmd","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":1563951785.0587456,"logger":"cmd","msg":"Version of operator-sdk: v0.8.0"}
{"level":"info","ts":1563951785.0590405,"logger":"leader","msg":"Trying to become the leader."}
{"level":"error","ts":1563951785.06033,"logger":"cmd","msg":"","error":"Get 
https://10.0.0.1:443/api?timeout=32s: dial tcp 10.0.0.1:443: 
connect: connection refused","stacktrace":"knative.dev/serving-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/mnt/data/gowork/src/knative.dev/serving-operator/vendor/github.com/go-logr/zapr/zapr.go:128\n
main.main\n\t/mnt/data/gowork/src/knative.dev/serving-operator/cmd/manager/main.go:97\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:200"}
```

yu2003w:
leader election is not used any more.

yu2003w:
Serving-Operator still restarted and reported following errors,
```
root@filmy1:/mnt/data/gowork/src/knative.dev/serving-operator# kubectl logs -f -c knative-serving-operator knative-serving-operator-7ff9bf4cd6-7zzqj
2019-08-08T08:07:00.112Z	ERROR	kubebuilder.manager	Failed to get API Group-Resources	{"error": "Get https://10.0.0.1:443/api?timeout=32s: dial tcp 10.0.0.1:443: connect: connection refused"}
knative.dev/serving-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error
	/mnt/data/gowork/src/knative.dev/serving-operator/vendor/github.com/go-logr/zapr/zapr.go:128
knative.dev/serving-operator/vendor/sigs.k8s.io/controller-runtime/pkg/manager.New
	/mnt/data/gowork/src/knative.dev/serving-operator/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go:173
main.main
[...]

PR #88, opened 2019-07-24
leader.Become() failed when kube-api couldn't be reached during startup

yu2003w:
Fixed:  #87

Added poll logic in case kube-api couldn't be reached when leader.Become() is called.

Signed-off-by: Wei Yu <yuweiw@cn.ibm.com>

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -21,8 +21,11 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+
+	"k8s.io/apimachinery/pkg/util/wait"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"knative.dev/serving-operator/pkg/apis"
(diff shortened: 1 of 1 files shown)


markusthoemmes:
I'd say we just ditch leader election for now. It's not something we use in Knative yet, it'll likely not be needed here as well.

cmd/manager/main.go
```diff
@@ -91,10 +94,15 @@ func main() {
 
 	ctx := context.TODO()
 
+	// We sometimes startup faster than we can reach kube-api. Poll on failure to prevent us terminating
 	// Become the leader before proceeding
-	err = leader.Become(ctx, "knative-serving-operator-lock")
-	if err != nil {
-		log.Error(err, "")
+	if perr := wait.PollImmediate(time.Second, 60*time.Second, func() (bool, error) {
```
houshengbo: @yu2003w I do not think we will need the leader selection any longer, coz the new controller model does not need it.

yu2003w: ok, I will close this. Thanks.

yu2003w:
Close this PR as it's not applicable for current implementation.

Issue #81, opened 2019-07-18
operator should ensure all serving components are remain installed as long as CR is present


yu2003w:
/kind feature

**Branch:** 
/master
/HEAD

**Current Behavior:**
Currently after installing Knative-serving with serving-operator, all components such as controller, activator, autoscaler and so on are deployed but not become 'available' due to some other reasons.
For example, image pull secret is missing.
Then delete the deployments in namespace knative-serving, operator will not reinstall the deployments again.

**Expected Behavior:**
operator could reinstall deployments as long as CR still present.

 **Detailed Procedure**
1, remove image pull secret from namespace knative-serving
2, update serving_v1alpha1_knativeserving_cr.yaml to apply customized images
```
spec:
  registry:
    override:
      activator: icpdev.icp:8500/imgrepo/serving-activator-amd64:v0.6.1
      autoscaler: icpdev.icp:8500/imgrepo/serving-autoscaler-amd64:v0.6.1
      controller: icpdev.icp:8500/imgrepo/serving-controller-amd64:v0.6.1
      networking-certmanager: icpdev.icp:8500/imgrepo/serving-certmanager-amd64:v0.6.1
      networking-istio: icpdev.icp:8500/imgrepo/serving-istio-amd64:v0.6.1
      webhook: icpdev.icp:8500/imgrepo/serving-webhook-amd64:v0.6.1
```
2, launch serving-operator to install knative-serving
3, delete deployments in namespace knative-serving.
deleted deployments are not created again.

savitaashture:
@yu2003w can you give me little more detail like where you run operator and which version.
Because i verified in openshift its working and operator able to recreate deployment if its deleted.

yu2003w:
@savitaashture I deployed serving-operator with master branch and deploy on original k8s cluster. I don't have a working openshift cluster to try now.

savitaashture:
@yu2003w to make sure i just did following steps
1. Installed minikube
2. Installed istio following https://github.com/knative/serving-operator#prerequisites
3. created
    ```
    kubectl apply -f config/crds/serving_v1alpha1_knativeserving_crd.yaml
    ko apply -f config/
    kubectl create -f config/crds/serving_v1alpha1_knativeserving_cr.yaml
    ```
4. All deployments running in `knative-serving` namespace
```
kubectl get pods -n knative-serving
NAME                                     READY   STATUS    RESTARTS   AGE
activator-5ccf59c56-6hrgc                1/1     Running   0          79s
[...]

PR #89, opened 2019-07-27
deployments should be created automatically as long as CR is present

yu2003w:
Fixed: #81 

mark instance status to installed failed if deployments are deleted manually after installed successfully

Signed-off-by: Wei Yu <yuweiw@cn.ibm.com>

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -252,7 +252,9 @@ func (r *ReconcileKnativeServing) checkDeployments(instance *servingv1alpha1.Kna
 			if err := r.client.Get(context.TODO(), key, deployment); err != nil {
 				instance.Status.MarkDeploymentsNotReady()
 				if errors.IsNotFound(err) {
-					return nil
+					// maybe deployments deleted manually
+					log.Info("checkDeployments", u.GetName(), " not found, maybe deleted manually")
+					instance.Status.MarkInstallFailed("deployments missed")
 				}
 				return err
 			}


yu2003w:
/assign @jcrossley3
@jcrossley3 Could you please help to have a look at this changes?

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -252,7 +252,9 @@ func (r *ReconcileKnativeServing) checkDeployments(instance *servingv1alpha1.Kna
 			if err := r.client.Get(context.TODO(), key, deployment); err != nil {
 				instance.Status.MarkDeploymentsNotReady()
 				if errors.IsNotFound(err) {
-					return nil
+					// maybe deployments deleted manually
+					log.Info("checkDeployments", u.GetName(), " not found, maybe deleted manually")
+					instance.Status.MarkInstallFailed("deployments missed")
```
houshengbo: @yu2003w I have resolved this issue in the PR 83 for the refactoring, by re-implementing this fuction..

markusthoemmes: I don't think this change is necessary. The `ApplyAll` will fix it anyway, will it?

jcrossley3: This code path will most often be taken in the time it takes kubernetes to spin up the deployments after the operator initially creates them from its manifest. For that pretty common race condition, I think this change may add more confusion than clarity. So it's not clear to me exactly what problem this is solving.

yu2003w: I installed kserving with operator by override registry images which resides in private docker registry. Because lack of image pull secretes, deployments are not ready when deployed.
Then I deleted two deployments webhook and controller.
Later I created registry secret and patched to proper service account. And deployment web hook and controller are not recreated.
So I want to fix this issue.
As the problem has been resolved in PR#83, I will close this PR.

yu2003w:
Close PR as it's fixed in #83.

PR #99, opened 2019-08-01
Cleanup vendored projects.

markusthoemmes:
This cleans up our vendor folder quite a bit as it drops dependencies wedon't actually need. It also replaces revisions with actual versions for better readability.


markusthoemmes:
/assign @houshengbo 

typo...

PR #101, opened 2019-08-01
Remove most of the operator-sdk specific functionality.

markusthoemmes:
`pkg/controller` does not do any of the things removed here or it does them different intrinsically.


houshengbo:
@markusthoemmes Generally looking good. I guess I dump all of them in the PR83.
I lean on to get the integration tests in first, so we know we are safe to remove.

markusthoemmes:
Note sure if this can cause any breakage but sure, I'm fine to wait for this to after the e2e tests.

The whole idea of creating the other PR though is to show that we don't need to "dump" it all into one PR but to cut them into small incremental changes, that we can PR (and thus review) separately.

cmd/manager/main.go
```diff
@@ -62,7 +57,7 @@ func main() {
 	// implementing the logr.Logger interface. This logger will
 	// be propagated through the whole operator, generating
 	// uniform and structured logs.
-	logf.SetLogger(zap.Logger())
+	logf.SetLogger(logf.ZapLogger(true))
```
houshengbo: `logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"` has been marked as deprecated https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/runtime/log/log.go, but I think we can take it by far, since it is part of the transition.

PR #105, opened 2019-08-01
Revert "Cleanup vendored projects. (#99)"

markusthoemmes:
I screwed something up and want to fix it properly. Let's revert to unblock though!


houshengbo:
It seems we have to slow down, in order to make any major progress, but still many thanks to @markusthoemmes for the trial.
/lgtm

markusthoemmes:
Yeah, let's do the more important changes first.

PR #108, opened 2019-08-02
Add test cases to verify the creation of KnativeServing 🧗‍♂️

houshengbo:
## Proposed Changes

* This PR adds the test case to create the knativeserving under a specified test namespace.
* Verify the deployment recreation under the test namespace.
* Verify the deletion of knativeserving.

**Release Note**

```release-note

```

### DIFF
--- a/test/cleanup.go
+++ b/test/cleanup.go
@@ -43,7 +43,4 @@ func TearDown(clients *Clients, names ResourceNames) {
 	if clients != nil && clients.KnativeServingAlphaClient != nil {
 		clients.KnativeServingAlphaClient.Delete(names.KnativeServing, &metav1.DeleteOptions{})
 	}
-
-	// Remove the namespace
-	clients.KubeClient.Kube.CoreV1().Namespaces().Delete(names.Namespace, &metav1.DeleteOptions{})
 }
--- a/test/cleanup.go
+++ b/test/cleanup.go
@@ -44,16 +44,6 @@ func TearDown(clients *Clients, names ResourceNames) {
 		clients.KnativeServingAlphaClient.KnativeServings.Delete(names.KnativeServing, &metav1.DeleteOptions{})
 	}
 
-	// Remove all the deployments
-	dpListSave, err := clients.KubeClient.Kube.AppsV1().Deployments(names.Namespace).List(metav1.ListOptions{})
-	if err == nil && len(dpListSave.Items) == 0 {
-		// Delete the deployments one by one.
-		for _, deployment := range dpListSave.Items {
-			clients.KubeClient.Kube.AppsV1().Deployments(deployment.Namespace).Delete(deployment.Name,
-				&metav1.DeleteOptions{})
-		}
-	}
-
 	// Remove the namespace
 	clients.KubeClient.Kube.CoreV1().Namespaces().Delete(names.Namespace, &metav1.DeleteOptions{})
 }
--- a/test/cleanup.go
+++ b/test/cleanup.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// cleanup allows you to define a cleanup function that will be executed
+// if your test is interrupted.
+
+package test
+
+import (
+	"os"
+	"os/signal"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CleanupOnInterrupt will execute the function cleanup if an interrupt signal is caught
+func CleanupOnInterrupt(cleanup func()) {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for range c {
+			cleanup()
+			os.Exit(1)
+		}
+	}()
+}
+
+// TearDown will delete created names using clients.
+func TearDown(clients *Clients, names ResourceNames) {
+	if clients != nil && clients.KnativeServingAlphaClient != nil {
+		clients.KnativeServingAlphaClient.KnativeServings.Delete(names.KnativeServing, &metav1.DeleteOptions{})
+	}
+
+	// Remove all the deployments
+	dpListSave, err := clients.KubeClient.Kube.AppsV1().Deployments(names.Namespace).List(metav1.ListOptions{})
+	if err == nil && len(dpListSave.Items) == 0 {
+		// Delete the deployments one by one.
+		for _, deployment := range dpListSave.Items {
+			clients.KubeClient.Kube.AppsV1().Deployments(deployment.Namespace).Delete(deployment.Name,
+				&metav1.DeleteOptions{})
+		}
+	}
+
+	// Remove the namespace
+	clients.KubeClient.Kube.CoreV1().Namespaces().Delete(names.Namespace, &metav1.DeleteOptions{})
+}
--- a/test/cleanup.go
+++ b/test/cleanup.go
@@ -41,7 +41,7 @@ func CleanupOnInterrupt(cleanup func()) {
 // TearDown will delete created names using clients.
 func TearDown(clients *Clients, names ResourceNames) {
 	if clients != nil && clients.KnativeServingAlphaClient != nil {
-		clients.KnativeServingAlphaClient.KnativeServings.Delete(names.KnativeServing, &metav1.DeleteOptions{})
+		clients.KnativeServingAlphaClient.Delete(names.KnativeServing, &metav1.DeleteOptions{})
 	}
 
 	// Remove the namespace
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -14,9 +14,11 @@ limitations under the License.
 package e2e
 
 import (
+	"k8s.io/apimachinery/pkg/util/wait"
 	"testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	// Mysteriously required to support GCP auth (required by k8s libs).
 	// Apparently just importing it is enough. @_@ side effects @_@.
 	// https://github.com/kubernetes/client-go/issues/242
@@ -25,6 +27,7 @@ import (
 	"knative.dev/serving-operator/test"
 	"knative.dev/serving-operator/test/resources"
 )
+
 
 // Setup creates the client objects needed in the e2e tests.
 func Setup(t *testing.T) *test.Clients {
@@ -69,11 +72,20 @@ func deploymentRecreation(t *testing.T, clients *test.Clients, names test.Resour
 			&metav1.DeleteOptions{}); err != nil {
 			t.Fatalf("Failed to delete deployment %s/%s: %v", deployment.Namespace, deployment.Name, err)
 		}
-		if _, err = resources.WaitForDeployment(clients, deployment.Name, deployment.Namespace,
-			resources.IsDeploymentAvailable, "DeploymentIsAvailable"); err != nil {
-			t.Fatalf("The deployment %s/%s failed to reach the desired state: %v",
-				deployment.Namespace, deployment.Name, err)
+
+		waitErr := wait.PollImmediate(resources.Interval, resources.Timeout, func() (bool, error) {
+			dep, err := clients.KubeClient.Kube.AppsV1().Deployments(deployment.Namespace).Get(deployment.Name, metav1.GetOptions{})
+			if err != nil && apierrs.IsNotFound(err) {
+				// If the deployment is not found, we continue to wait for the availability.
+				return false, nil
+			}
+			return resources.IsDeploymentAvailable(dep)
+		})
+
+		if waitErr != nil {
+			t.Fatalf("The deployment %s/%s failed to reach the desired state: %v", deployment.Namespace, deployment.Name, err)
 		}
+
 		if _, err := resources.WaitForKnativeServingState(clients.KnativeServingAlphaClient, test.ServingOperatorName,
 			resources.IsKnativeServingReady); err != nil {
 			t.Fatalf("KnativeService %q failed to reach the desired state: %v", test.ServingOperatorName, err)
@@ -98,10 +110,17 @@ func knativeServingDeletion(t *testing.T, clients *test.Clients, names test.Reso
 	}
 
 	for _, deployment := range dpListSave.Items {
-		if _, err := resources.WaitForDeployment(clients, deployment.Name, deployment.Namespace,
-			resources.IsDeploymentDeleted, "DeploymentIsDeleted"); err != nil {
-			t.Fatalf("The deployment %s/%s failed to be deleted: %v",
-				deployment.Namespace, deployment.Name, err)
+		waitErr := wait.PollImmediate(resources.Interval, resources.Timeout, func() (bool, error) {
+			if _, err := clients.KubeClient.Kube.AppsV1().Deployments(deployment.Namespace).Get(deployment.Name,
+				metav1.GetOptions{}); err != nil && apierrs.IsNotFound(err) {
+				// If the deployment is not found, we determine it is deleted.
+				return true, nil
+			}
+			return false, nil
+		})
+
+		if waitErr != nil {
+			t.Fatalf("The deployment %s/%s failed to be deleted: %v", deployment.Namespace, deployment.Name, err)
 		}
 		t.Logf("The deployment %s/%s has been deleted.", deployment.Namespace, deployment.Name)
 	}
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -24,7 +24,6 @@ import (
 	"knative.dev/serving-operator/test"
 )
 
-
 // Setup creates the client objects needed in the e2e tests.
 func Setup(t *testing.T) *test.Clients {
 	return SetupWithNamespace(t, test.ServingOperatorNamespace)
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -14,18 +14,14 @@ limitations under the License.
 package e2e
 
 import (
-	"k8s.io/apimachinery/pkg/util/wait"
 	"testing"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	// Mysteriously required to support GCP auth (required by k8s libs).
 	// Apparently just importing it is enough. @_@ side effects @_@.
 	// https://github.com/kubernetes/client-go/issues/242
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	pkgTest "knative.dev/pkg/test"
 	"knative.dev/serving-operator/test"
-	"knative.dev/serving-operator/test/resources"
 )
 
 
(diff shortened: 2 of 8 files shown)


markusthoemmes:
This seems like a duplicate of #106?

houshengbo:
@markusthoemmes Right, I will accept that one and proceed with adding the KnativeServing test case.

houshengbo:
Plz remove the logic here: https://github.com/knative/serving-operator/blob/master/test/e2e/knativeservingdeployment_test.go#L38, and the other test file https://github.com/knative/serving-operator/blob/master/test/e2e/knativeserving_test.go.

test/cleanup.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// TearDown will delete created names using clients.
+func TearDown(clients *Clients, names ResourceNames) {
+	if clients != nil && clients.KnativeServingAlphaClient != nil {
+		clients.KnativeServingAlphaClient.KnativeServings.Delete(names.KnativeServing, &metav1.DeleteOptions{})
+	}
+
+	// Remove all the deployments
+	dpListSave, err := clients.KubeClient.Kube.AppsV1().Deployments(names.Namespace).List(metav1.ListOptions{})
```
jcrossley3: This isn't necessary. Just deleting the KS instance should suffice

test/e2e/e2e.go
```diff
@@ -40,3 +45,73 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+			t.Fatalf("The deployment %s/%s failed to reach the desired state: %v",
+				deployment.Namespace, deployment.Name, err)
+		}
+		if _, err := resources.WaitForKnativeServingState(clients.KnativeServingAlphaClient, test.ServingOperatorName,
+			resources.IsKnativeServingReady); err != nil {
+			t.Fatalf("KnativeService %q failed to reach the desired state: %v", test.ServingOperatorName, err)
+		}
+		t.Logf("The deployment %s/%s reached the desired state.", deployment.Namespace, deployment.Name)
+	}
+	return dpListSave
```
jcrossley3: I wouldn't return this. Let the tests be independent. Don't pass a return from one to another.

test/e2e/knativeservingdeployment_test.go
```diff
@@ -18,38 +18,37 @@ package e2e
...
+	test.CleanupOnInterrupt(func() { test.TearDown(clients, names) })
+	defer test.TearDown(clients, names)
+
+	// Create the namespace for tests
+	CreateNamespace(t, clients, names.Namespace)
+
+	// Change the namespace for the clients
+	clients = SetupWithNamespace(t, names.Namespace)
+
+	// Create a KnativeServing to see if it can reach the READY status
```
jcrossley3: I still think you should turn these into "subtests" so that they can be run independently (to the extent possible). It'll give better status output when running them locally, and we should be able to make it so that you can pass a `-run` argument to run single tests. Something along these lines:

```
	CreateKnativeServing(t, clients, names)
	t.Run("create", func(t *testing.T) {
		// Create a KnativeServing to see if it can reach the READY status
		TestKnativeServingCreation(t, clients, names)
	})
	t.Run("restore", func(t *testing.T) {
		// Delete the deployments one by one to see if they will be recreated.
		ensureReady()
		TestDeploymentRecreation(t, clients, names)
	})
	t.Run("delete", func(t *testing.T) {
		// Delete the KnativeServing to see if all the deployments will be removed as well
		ensureReady()
		TestKnativeServingDeletion(t, clients, names)
	})

```

test/resources/knativeserving.go
```diff
@@ -36,20 +38,78 @@ import (
...
-	timeout  = 2 * time.Minute
+	timeout  = 5 * time.Minute
 )
 
+// CreateKnativeServingReady polls the status of the KnativeServing called name
+// from client every `interval` until `inState` returns `true` indicating it
+// is done, returns an error or timeout.
+func CreateKnativeServingReady(t *testing.T, clients *test.KnativeServingAlphaClients, names test.ResourceNames,
+	inState func(s *v1alpha1.KnativeServing, err error) (bool, error)) (*v1alpha1.KnativeServing, error) {
+	svc, err := CreateKnativeServing(t, clients, names)
```
jcrossley3: I would move this out of here. Since all the tests require a KS, just have this test verify a Ready state, assuming the KS has already been created (see above).

test/resources/knativeserving.go
```diff
@@ -36,20 +38,78 @@ import (
...
+	if waitErr != nil {
+		return lastState, errors.Wrapf(waitErr, "knativeserving %s is not in desired state, got: %+v", names.KnativeServing, lastState)
+	}
+	return lastState, nil
+}
+
+func getDefaultKSSpec() *v1alpha1.KnativeServingSpec {
+	return &v1alpha1.KnativeServingSpec{
+		Config: map[string]map[string]string{
+			"defaults": {
```
jcrossley3: If we're going to include any config here, we should verify that the operator updates the correct configmap[s], so choose a value that wouldn't be a default.

test/resources/knativeserving.go
```diff
@@ -59,22 +119,28 @@ func WaitForKnativeServingState(client *test.KnativeServingAlphaClients, name st
...
+// IsKnativeServingDeleted will check the status conditions of the KnativeServing and return true if the KnativeServing is deleted.
+func IsKnativeServingDeleted(s *v1alpha1.KnativeServing, err error) (bool, error) {
+	return apierrs.IsNotFound(err), nil
+}
+
+// WaitForDeployment polls the status of the deployment called name
 // from client every `interval` until `inState` returns `true` indicating it
 // is done, returns an error or timeout.
-func WaitForDeploymentAvailable(clients *test.Clients, name, namespace string, inState func(s *v1.Deployment) (bool, error)) (*v1.Deployment, error) {
+func WaitForDeployment(clients *test.Clients, name, namespace string, inState func(s *v1.Deployment, err error) (bool, error),
```
jcrossley3: Why not use the upstream `pkg/test/WaitForDeploymentState`?

test/cleanup.go
houshengbo: OK.

test/e2e/e2e.go
houshengbo: since we do not need to delete the deployments in the teardown, we do not need the return any mor.

test/resources/knativeserving.go
houshengbo: @jcrossley3  The problem I had here is that if I only defined the name and namespace, KS would never go to the ready status, and no deployment would be created.

test/resources/knativeserving.go
houshengbo: good catch. will use it.

test/e2e/e2e.go
```diff
@@ -40,3 +45,73 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
 	}
 	return clients
 }
+
+// CreateNamespace creates a namespace to run the integration tests.
+func CreateNamespace(t *testing.T, clients *test.Clients, namespace string) {
```
markusthoemmes: Please do not do this in code. It prevents you from modifying the namespaces in third-party integration systems should you need it. It's also different from what knative-serving does, where the namespaces are created from `test/config`

test/e2e/e2e.go
```diff
@@ -40,3 +45,73 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+// TestKnativeServingCreation creates a KnativeServing to see if it can reach the READY status.
+func TestKnativeServingCreation(t *testing.T, clients *test.Clients, names test.ResourceNames) {
+	if _, err := resources.CreateKnativeServingReady(t, clients.KnativeServingAlphaClient, names,
+		resources.IsKnativeServingReady); err != nil {
+		t.Fatalf("KnativeService %q failed to get to the READY status: %v", names.KnativeServing, err)
+	}
+}
+
+// TestDeploymentRecreation verify whether all the deployments for knative serving are able to recreate, when they are deleted.
+func TestDeploymentRecreation(t *testing.T, clients *test.Clients, names test.ResourceNames) *v1.DeploymentList {
```
markusthoemmes: Let's please not name functions `TestXXX` if they are not actually tests.

test/resources/knativeserving.go
markusthoemmes: I believe setting a pointer to an empty spec should work. FWIW, we should make it that an `nil` pointer does the right thing (use defaults) as well. This is a bug if it doesn't.

test/e2e/e2e.go
```diff
@@ -40,3 +45,73 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+		// Create the namespace if not available
+		if _, err = clients.KubeClient.Kube.CoreV1().Namespaces().Create(&corev1.Namespace{ObjectMeta:
+			metav1.ObjectMeta{Name: namespace}}); err != nil {
+			t.Fatalf("Failed to create the namespace %s: %v", namespace, err)
+		}
+	}
+}
+
+// TestKnativeServingCreation creates a KnativeServing to see if it can reach the READY status.
+func TestKnativeServingCreation(t *testing.T, clients *test.Clients, names test.ResourceNames) {
```
markusthoemmes: Why are these functions big top-level package test helpers vs. just helpers in the test file itself?

test/resources/knativeserving.go
jcrossley3: Tests pass for me locally after deleting `Spec: *getDefaultKSSpec(),` from the `KnativeServing` instance, so I think we're good. I still think it's useful to test with/without config and ensure the corresponding configmaps are updated, though.

test/e2e/e2e.go
jcrossley3: Another benefit, at least of not having a randomized namespace name, is that once we have subtests we can run them individually, e.g.
```
[jim@t450s serving-operator]$ go test -run=/create -v -tags=e2e -count=1 ./test/e2e
=== RUN   TestKnativeServingDeployment
=== RUN   TestKnativeServingDeployment/create
--- PASS: TestKnativeServingDeployment (50.12s)
    --- PASS: TestKnativeServingDeployment/create (50.02s)
PASS
ok  	knative.dev/serving-operator/test/e2e	50.154s
[jim@t450s serving-operator]$ go test -run=/delete -v -tags=e2e -count=1 ./test/e2e
=== RUN   TestKnativeServingDeployment
=== RUN   TestKnativeServingDeployment/delete
--- FAIL: TestKnativeServingDeployment (0.09s)
    --- FAIL: TestKnativeServingDeployment/delete (0.02s)
        e2e.go:105: KnativeService "knative-serving" failed to delete: knativeservings.serving.knative.dev "knative-serving" not found
FAIL
FAIL	knative.dev/serving-operator/test/e2e	0.122s
```
It's the randomized namespace that causes the 2nd failure.

test/e2e/e2e.go
houshengbo: I renamed them

test/cleanup.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// TearDown will delete created names using clients.
+func TearDown(clients *Clients, names ResourceNames) {
+	if clients != nil && clients.KnativeServingAlphaClient != nil {
+		clients.KnativeServingAlphaClient.KnativeServings.Delete(names.KnativeServing, &metav1.DeleteOptions{})
+	}
+
+	// Remove the namespace
+	clients.KubeClient.Kube.CoreV1().Namespaces().Delete(names.Namespace, &metav1.DeleteOptions{})
```
jcrossley3: Don't do this. Whoever creates the ns should delete it.

test/e2e/e2e.go
```diff
@@ -40,3 +44,79 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+		// Create the namespace if not available
+		if _, err = clients.KubeClient.Kube.CoreV1().Namespaces().Create(&corev1.Namespace{ObjectMeta:
+			metav1.ObjectMeta{Name: namespace}}); err != nil {
+			t.Fatalf("Failed to create the namespace %s: %v", namespace, err)
+		}
+	}
+}
+
+// KnativeServingVerify verifies if the KnativeServing can reach the READY status.
+func KnativeServingVerify(t *testing.T, clients *test.Clients, names test.ResourceNames) {
```
jcrossley3: nit, but do any of these really need to be public?

test/resources/knativeserving.go
```diff
@@ -36,20 +38,53 @@ import (
...
+	}
+	svc, err := clients.KnativeServings.Create(ks)
+	return svc, err
+}
+
 // WaitForKnativeServingState polls the status of the KnativeServing called name
 // from client every `interval` until `inState` returns `true` indicating it
 // is done, returns an error or timeout.
-func WaitForKnativeServingState(client *test.KnativeServingAlphaClients, name string, inState func(s *v1alpha1.KnativeServing) (bool, error)) (*v1alpha1.KnativeServing, error) {
+func WaitForKnativeServingState(client *test.KnativeServingAlphaClients, name string, inState func(s *v1alpha1.KnativeServing,
```
jcrossley3: i'm curious why you don't just pass `servingv1alpha1.KnativeServingInterface` in as the client. What's the value of wrapping it in `test.KnativeServingAlphaClients`?

test/resources/knativeserving.go
jcrossley3: I don't understand. You seem to be adding the error arg, so that would enable you to use the existing one, right?

houshengbo: After adding error param, cannot use the existing one in pkg/test/WaitForDeploymentState.

test/e2e/e2e.go
houshengbo: they can be private so far.

test/resources/knativeserving.go
houshengbo: So far, it is the only resource KnativeServing that we will access. KnativeServingAlphaClients is designed to support multiple clients accessing multiple resources.
I am simply it.

test/resources/knativeserving.go
jcrossley3: Gotcha. So that's even easier. Your `inState` can just always return false, because it won't even get called at all if everything works as expected. You'll just need to do your `IsNotFound` check in the test calling `WaitForDeploymentState` because it will be returning the error you expect.

houshengbo: I use this func WaitForDeployment in two situations: one to wait until the deployment is available, the other is to wait until it is removed. For the first one I need to check the deployment itself, but for the second one I need to check if error is NotFound, indicating deployment is gone.  For the instate func to verify the status, I just take both of the deployment and error as params, so WaitForDeployment can resued for these two cases. It has to be different from pkg/test/WaitForDeploymentState.

jcrossley3: I understand. All I'm suggesting is that you can simplify our code by using the existing `pkg/test` helper, removing the `error` arg from your instate functions, having `isDeploymentDeleted` always return false and changing your fatal failure condition in `TestKnativeServingDeletion` to something like this:
```
	for _, deployment := range dpListSave.Items {
		if _, err := test.WaitForDeploymentState(..., resources.IsDeploymentDeleted, ...); !apierrs.IsNotFound(err) {
			t.Fatalf(...)
		}
		t.Logf(...)
```
Make sense?

houshengbo: if resources.IsDeploymentDeleted always return false, WaitForDeploymentState will never get the desired condition, before timeout. 

when IsDeploymentDeleted return true, WaitForDeploymentState meet the desired condition to quite.

houshengbo: I think WaitForDeploymentState can be used to check if deployment is removed,
but for waiting deployment till available, WaitForDeploymentState will return error, if deployment is not found.
I still do not know how to use it for this case.
During deployment recreation, it is possible that deployment is not available, but will come back later.

test/e2e/e2e.go
```diff
@@ -40,3 +45,83 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+func knativeServingVerify(t *testing.T, clients *test.Clients, names test.ResourceNames) {
+	if _, err := resources.WaitKnativeServingReady(t, clients.KnativeServingAlphaClient, names.KnativeServing,
+		resources.IsKnativeServingReady); err != nil {
+		t.Fatalf("KnativeService %q failed to get to the READY status: %v", names.KnativeServing, err)
+	}
+
+}
+
+// deploymentRecreation verify whether all the deployments for knative serving are able to recreate, when they are deleted.
+func deploymentRecreation(t *testing.T, clients *test.Clients, names test.ResourceNames) {
```
markusthoemmes: Do you anticipate us ever reusing these helpers? I feel like they belong next to the test in the appropriate file.

test/e2e/e2e.go
```diff
@@ -40,3 +45,83 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+		t.Logf("The deployment %s/%s reached the desired state.", deployment.Namespace, deployment.Name)
+	}
+}
+
+// knativeServingDeletion deletes tha KnativeServing to see if all the deployments will be removed.
+func knativeServingDeletion(t *testing.T, clients *test.Clients, names test.ResourceNames) {
+	if err := clients.KnativeServingAlphaClient.Delete(names.KnativeServing, &metav1.DeleteOptions{}); err != nil {
+		t.Fatalf("KnativeService %q failed to delete: %v", names.KnativeServing, err)
+	}
+	if _, err := resources.WaitForKnativeServingState(clients.KnativeServingAlphaClient, names.KnativeServing,
```
markusthoemmes: Why do we need to wait for this state? Is this delayed because of finalizers? I'd think the above would block until it's really really deleted so checking its error might suffice?

test/e2e/e2e.go
```diff
@@ -40,3 +45,83 @@ func SetupWithNamespace(t *testing.T, namespace string) *test.Clients {
...
+
+	dpListSave, err := clients.KubeClient.Kube.AppsV1().Deployments(names.Namespace).List(metav1.ListOptions{})
+	if err != nil {
+		t.Fatalf("Error getting any deployment under the namespace %q: %v", names.Namespace, err)
+	}
+
+	for _, deployment := range dpListSave.Items {
+		waitErr := wait.PollImmediate(resources.Interval, resources.Timeout, func() (bool, error) {
+			if _, err := clients.KubeClient.Kube.AppsV1().Deployments(deployment.Namespace).Get(deployment.Name,
+				metav1.GetOptions{}); err != nil && apierrs.IsNotFound(err) {
```
markusthoemmes: We're ignoring other errors here, which might cause the tests to fail for bogus reasons. I'd envision the logic to look more like:

```golang
_, err := clients.KubeClient.Kube.AppsV1().Deployments(deployment.Namespace).Get(deployment.Name, metav1.GetOptions{})
if err != nil {
  if apierrs.IsNotFound(err) {
    return true, nil
  }
  return false, err
}
return false, nil
```

That'll cause other errors (like a wrongly setup test env) to be surfaced properly.

test/resources/knativeserving.go
```diff
@@ -21,68 +21,88 @@ package resources
...
+	// Timeout specifies the timeout for the function PollImmediate to reach a certain status.
+	Timeout  = 5 * time.Minute
 )
 
-// WaitForKnativeServingState polls the status of the KnativeServing called name
+// WaitKnativeServingReady polls the status of the KnativeServing called name
 // from client every `interval` until `inState` returns `true` indicating it
 // is done, returns an error or timeout.
-func WaitForKnativeServingState(client *test.KnativeServingAlphaClients, name string, inState func(s *v1alpha1.KnativeServing) (bool, error)) (*v1alpha1.KnativeServing, error) {
+func WaitKnativeServingReady(t *testing.T, clients servingv1alpha1.KnativeServingInterface, name string,
```
markusthoemmes: Why rename this? You still have an inState function there that might check for different things than "ready"?

test/resources/knativeserving.go
```diff
@@ -21,68 +21,88 @@ package resources
...
 }
 
-// WaitForDeploymentAvailable polls the status of the deployment called name
+// WaitForKnativeServingState polls the status of the KnativeServing called name
 // from client every `interval` until `inState` returns `true` indicating it
 // is done, returns an error or timeout.
-func WaitForDeploymentAvailable(clients *test.Clients, name, namespace string, inState func(s *v1.Deployment) (bool, error)) (*v1.Deployment, error) {
-	span := logging.GetEmitableSpan(context.Background(), fmt.Sprintf("WaitForKnativeServingState/%s/%s", name,
-		"DeploymentIsAvailable"))
+func WaitForKnativeServingState(client servingv1alpha1.KnativeServingInterface, name string, inState func(s *v1alpha1.KnativeServing,
```
markusthoemmes: How does this differ from `WaitKnativeServingReady` above?

test/resources/knativeserving.go
```diff
@@ -21,68 +21,88 @@ package resources
...
-		return dep, errors.Wrapf(waitErr, "Deployment %q is not in desired status for the condition type Available,"+
-			"got: %+q; want %+q", name, getDeploymentStatus(dep), "True")
+		return lastState, errors.Wrapf(waitErr, "knativeserving %q is not in desired state, got: %+v", name, lastState)
 	}
+	return lastState, nil
+}
+
+// IsKnativeServingReady will check the status conditions of the KnativeServing and return true if the KnativeServing is ready.
+func IsKnativeServingReady(s *v1alpha1.KnativeServing, err error) (bool, error) {
+	return s.Status.IsReady(), nil
```
markusthoemmes: I feel like these helpers should surface unexpected errors, i.e. return

```golang
s.Status.IsReady(), err
```

here and

```
if !apierrs.IsNotFound(err) {
  return false, err
}
return true, nil
```

below.

test/e2e/e2e.go
houshengbo: OK. If the resource is removed before clients.KnativeServingAlphaClient.Delete returns, then we do not need to verify the status of WaitForKnativeServingState.

test/e2e/e2e.go
houshengbo: Done

test/resources/knativeserving.go
houshengbo: The reason was I was not sure about the status during deletion, also verified the case of deletion. Since we do not need to verify for delete, I can change it back.

test/resources/knativeserving.go
houshengbo: Combine them now.

test/resources/knativeserving.go
houshengbo: No need of the func IsKnativeServingDeleted anymore, so I can remove it. Will change IsKnativeServingReady

test/e2e/e2e.go
houshengbo: I will move them into the test file.

test/e2e/e2e.go
```diff
@@ -24,6 +24,7 @@ import (
 	"knative.dev/serving-operator/test"
 )
 
+
```
markusthoemmes: Drop this diff please.

test/e2e/knativeservingdeployment_test.go
```diff
@@ -18,38 +18,128 @@ package e2e
...
+	if _, err := resources.WaitKnativeServingReady(clients.KnativeServingAlphaClient, names.KnativeServing,
+		resources.IsKnativeServingReady); err != nil {
+		t.Fatalf("KnativeService %q failed to get to the READY status: %v", names.KnativeServing, err)
+	}
+
+}
+
+// deploymentRecreation verify whether all the deployments for knative serving are able to recreate, when they are deleted.
+func deploymentRecreation(t *testing.T, clients *test.Clients, names test.ResourceNames) {
+	dpListSave, err := clients.KubeClient.Kube.AppsV1().Deployments(names.Namespace).List(metav1.ListOptions{})
```
markusthoemmes:
```suggestion
	dpList, err := clients.KubeClient.Kube.AppsV1().Deployments(names.Namespace).List(metav1.ListOptions{})
```

test/resources/knativeserving.go
```diff
@@ -27,62 +27,55 @@ import (
...
+	Timeout  = 5 * time.Minute
 )
 
-// WaitForKnativeServingState polls the status of the KnativeServing called name
+// WaitKnativeServingReady polls the status of the KnativeServing called name
 // from client every `interval` until `inState` returns `true` indicating it
 // is done, returns an error or timeout.
-func WaitForKnativeServingState(client *test.KnativeServingAlphaClients, name string, inState func(s *v1alpha1.KnativeServing) (bool, error)) (*v1alpha1.KnativeServing, error) {
-	span := logging.GetEmitableSpan(context.Background(), fmt.Sprintf("WaitForKnativeServingState/%s/%s", name, "KnativeServingIsReady"))
+func WaitKnativeServingReady(clients servingv1alpha1.KnativeServingInterface, name string,
```
markusthoemmes: Why is this called `WaitKnativeServingReady`? We are passing an instate function in still?

PR #109, opened 2019-08-02
Light cleanup on our dependencies.

markusthoemmes:
## Proposed Changes

A much more tuned down go at cleaning up our dependencies. This time only removing tools we really don't need and that are guaranteed to have no references.

I recreated `Gopkg.lock` from scratch in the last commit to guard against the breakage I caused the last time round. This looks much cleaner now!

/assign @houshengbo


trshafer:
Do all of the `hack` commands still work?

markusthoemmes:
@trshafer I think the bigboy is `update-codegen.sh` and yes, that still works a charm.

PR #112, opened 2019-08-02
Simplify extension logic in anticipation of ripping it out entirely

jcrossley3:
We'll leave the minikube extension in for now, just as a reminder that
we need something, but openshift is gone.

I change the args a bit, and they'll likely evolve... or not, if we move all extension stuff out of the repo.

See https://docs.google.com/document/d/1OtlkJ8whTgHYTz5rz0lq5mNdDGK_pakG/edit

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -186,47 +186,42 @@ func (r *ReconcileKnativeServing) updateStatus(instance *servingv1alpha1.Knative
 	return nil
 }
 
-// Apply the embedded resources
+// Install the resources from the Manifest
 func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServing) error {
 	log.V(1).Info("install", "status", instance.Status)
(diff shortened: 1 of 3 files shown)


pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -196,13 +196,7 @@ func (r *ReconcileKnativeServing) install(instance *servingv1alpha1.KnativeServi
...
 	defer r.updateStatus(instance)
 
-	extensions, err := platforms.Extend(r.client, r.kubeClientSet, r.dynamicClientSet, r.scheme)
-	if err != nil {
-		return err
-	}
-
-	err = r.executeInstall(extensions, instance)
-
+	err := r.executeInstall(instance)
```
trshafer: Personal preference: extract the transform into a separate function, and extract the status update to a separate function so the install function tells the story of
transform
applyAll
updateStatus

It would be totally fine to inline as well. But I personally prefer more small functions.

jcrossley3: done, thx

PR #127, opened 2019-08-08
Remove openshift specific code.

markusthoemmes:
## Proposed Changes

Remove pure openshift code for now as we're figuring out how to deal with different platforms.


jcrossley3:
This is subsumed by #112

PR #128, opened 2019-08-08
Drop the dependency on operator-sdk 🧗‍♂️

houshengbo:
Co-authored-by: Markus Thömmes \<markusthoemmes@me.com\>

## Proposed Changes

* This PR removes the dependency on operator-sdk.
* All the dependencies are regenerated by ./hack/update-deps.sh.

**Release Note**

```release-note

```


markusthoemmes:
I'd really punt on this for now. It's not necessary to continue the work on moving our code over. Especially the copying of the DynamicRestMapper file I did is a bit shady.

We can do this once we eliminated all usages of client.Client in the operator more easily and with less shady license tactics.

markusthoemmes:
Let's not do this yet.

/hold

houshengbo:
@markusthoemmes Any reason for that?

markusthoemmes:
See my comment above. We're copying code we shouldn't copy. Let's wait with this until we can actually drop that code.

markusthoemmes:
This is fine in this state now, sorry for the noise

/hold cancel

PR #133, opened 2019-08-09
Remove 'scheme'.

markusthoemmes:
## Proposed Changes

The `scheme` is used nowhere in our code today and I'd keep it that way.


houshengbo:
@markusthoemmes knative-prow-robot reminds you about the func comments.

houshengbo:
/lgtm
we can add comments later.

PR #146, opened 2019-08-16
Final refactoring PR for knative/pkg transition

houshengbo:
## Proposed Changes

* The sharedmain package is used for the new main function
* Dependencies are regenerated
* Log issue is fixed for manifestival
* Remove controller-runtime dependency
* Add yaml files for logging and observability

**Release Note**

```release-note

```


config/operator.yaml
```diff
@@ -26,3 +26,25 @@ spec:
...
+            - name: METRICS_DOMAIN
+              value: knative.dev/serving-operator
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: config-observability
+          volumeMounts:
+            - name: config-logging
+              mountPath: /etc/config-logging
+            - name: config-observability
+              mountPath: /etc/config-observability
+      volumes:
```
markusthoemmes: These volumes are no longer required.

pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -0,0 +1,264 @@
...
+
+}
+
+// Check that our Reconciler implements controller.Reconciler
+var _ controller.Reconciler = (*Reconciler)(nil)
+
+// Reconcile compares the actual state with the desired, and attempts to
+// converge the two. It then updates the Status block of the Knativeserving resource
+// with the current status of the resource.
+func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
```
markusthoemmes: Why are we creating a full blown copy of this? Makes diffing this against the old reconciler impossible.

pkg/reconciler/newreconciler/reconciler.go
```diff
@@ -0,0 +1,120 @@
...
+	sharedclientset "knative.dev/pkg/client/clientset/versioned"
+	sharedclient "knative.dev/pkg/client/injection/client"
+	servingclient "knative.dev/serving-operator/pkg/client/injection/client"
+	"knative.dev/pkg/configmap"
+	clientset "knative.dev/serving-operator/pkg/client/clientset/versioned"
+	servingScheme "knative.dev/serving-operator/pkg/client/clientset/versioned/scheme"
+)
+
+// Base implements the core controller logic, given a Reconciler.
+type Base struct {
```
markusthoemmes: This looks like a 1:1 copy of `pkg/reconciler/Base`. Do we need to copy it?

pkg/reconciler/knativeserving/controller.go
```diff
@@ -0,0 +1,86 @@
...
+	rbase "knative.dev/serving-operator/pkg/reconciler/newreconciler"
+)
+
+const (
+	controllerAgentName = "knativeserving-controller"
+	reconcilerName = "KnativeServing"
+)
+
+var (
+	ClusterConfig *rest.Config
```
markusthoemmes: We shouldn't use globals which we update at runtime, especially not in pkg. Can we not pass this value into the controller itself? Or just refetch it inside of the controller if that's not easily possible.

houshengbo:
@greghaynes @jcrossley3 @k4leung4 Shall we merge this one, since it has been no objection so far?

config/config-logging.yaml
```diff
@@ -0,0 +1,70 @@
+# Copyright 2018 The Knative Authors
```
k4leung4: nit: 2019

config/config-observability.yaml
```diff
@@ -0,0 +1,100 @@
+# Copyright 2018 The Knative Authors
```
k4leung4: nit: 2019

cmd/manager/main.go
```diff
@@ -17,72 +17,20 @@ package main
...
-
-	log.Info("Starting the Cmd.")
-
-	// Start the Cmd
-	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
-		log.Error(err, "Manager exited non-zero")
-		os.Exit(1)
+		log.Fatal("Error building kubeconfig", err)
 	}
+	sharedmain.MainWithConfig(signals.NewContext(), "controller", cfg, knativeserving.NewController)
```
greghaynes: I dont think we can should "controller" as the component name as this is already taken (although in another namespace). I also worry that the name is not really specific enough for the default namespace (pretty much anything can be called controller in the deafault ns). Can we use something like `servingoperator`

houshengbo: @greghaynes Done.

PR #159, opened 2019-09-04
Only delete shared resources when last KS is deleted

### LINKED ISSUE #154: What does it mean to have multiple KnativeServing CR's?
This is related to both #39 and #59 

Currently, when the KS CR is deleted, everything in the upstream manifest is deleted, the ns-scoped ones by the k8s garbage collector and the cluster-scoped ones by the operator. Technically, the operator will attempt to delete the ns-scoped ones, too, gracefully handling the race condition with the GC.

This means if there are more than one KS CR in the cluster, deleting any one of them will cause all of the cluster-scoped resources shared by the others to be deleted.

We need to decide if it's meaningful to have multiple KS CR's. I suspect that until k-s itself supports multi-tenancy, it isn't. And therefore, a simple solution would be for the operator to set any CR's status to an error condition if one KS CR already exists.

But if we do want to have multiple, the operator should probably support some form of reference counting to know when it's ok to delete shared, cluster-scoped resources.

jcrossley3:
Fixes #154 

Strictly speaking, it doesn't address what it means to have multiple `KnativeServing` resources, because having them isn't necessarily a problem. The problem occurs when any one of them is deleted, at which point (prior to this change) all the cluster-scoped resources are deleted, thereby breaking all the remaining KS installations.

With this fix, the shared resources are only deleted when the last KS is deleted.

### DIFF
--- a/pkg/reconciler/knativeserving/controller.go
+++ b/pkg/reconciler/knativeserving/controller.go
@@ -55,6 +55,7 @@ func NewController(
 	c := &Reconciler{
 		Base:                 rbase.NewBase(ctx, controllerAgentName, cmw),
 		knativeServingLister: knativeServingInformer.Lister(),
+		servings:             map[string]bool{},
 	}
 
 	koDataDir := os.Getenv("KO_DATA_PATH")
--- a/pkg/reconciler/knativeserving/controller.go
+++ b/pkg/reconciler/knativeserving/controller.go
@@ -22,6 +22,7 @@ import (
 	"github.com/go-logr/zapr"
 	mf "github.com/jcrossley3/manifestival"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	"knative.dev/pkg/configmap"
@@ -55,7 +56,7 @@ func NewController(
 	c := &Reconciler{
 		Base:                 rbase.NewBase(ctx, controllerAgentName, cmw),
 		knativeServingLister: knativeServingInformer.Lister(),
-		servings:             map[string]bool{},
+		servings:             sets.String{},
 	}
 
 	koDataDir := os.Getenv("KO_DATA_PATH")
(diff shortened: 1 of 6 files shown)


pkg/reconciler/knativeserving/controller.go
```diff
@@ -55,6 +55,7 @@ func NewController(
 	c := &Reconciler{
 		Base:                 rbase.NewBase(ctx, controllerAgentName, cmw),
 		knativeServingLister: knativeServingInformer.Lister(),
+		servings:             map[string]bool{},
```
evankanderson: Do you want to use a [`sets.String`](https://godoc.org/k8s.io/apimachinery/pkg/util/sets#String) here? It's not a lot more space to store the bool, but then you have to wonder if there's a meaning to `true` and `false`.

jcrossley3: Good call. Done.

houshengbo:
@jcrossley3 The integration test needs to be changed accordingly, though it does not fail so far. 

We need to check if all the KS have been removed, before verifying the cluster-scoped resource:
https://github.com/knative/serving-operator/blob/master/test/e2e/knativeservingdeployment_test.go#L156

jcrossley3:
> @jcrossley3 The integration test needs to be changed accordingly, though it does not fail so far.

I debated changing the test, but after verifying that the test actually does fail if your cluster already has a KS, I decided against it. Do you feel strongly that it's worth the effort to validate a meaningless scenario?

jcrossley3:
@houshengbo Check out the latest commit to see if that works better for you. I refactored `Clients` a bit to access all namespaces.

test/e2e/e2e.go
```diff
@@ -26,15 +26,14 @@ import (
 
 // Setup creates the client objects needed in the e2e tests.
 func Setup(t *testing.T) *test.Clients {
-	return SetupWithNamespace(t, test.ServingOperatorNamespace)
+	return SetupWithNamespace(t)
```
houshengbo: Could you consolidate the function SetupWithNamespace with Setup?

jcrossley3: Good call. Done.

Issue #163, opened 2019-09-10
KnativeServing operator should enforce the configuration defined in the CR


garron:
This issue was split out of #138 to separate the internal operator state management (handled in that issue) from discussion on how the operator should go about ensuring that it's configuration is in effect.

Consensus at the operator working group meeting was that the KnativeServing CR should be the source of truth and this issue is intended to cover work in making that so.

duglin:
I was going to open an issue but I think this one might be related, if not let me know and I'll open a new one.

Overall: where should a user modify things if they want to change their Knative install?

Today, w/o the operator, if someone wants to do something like modify the template of the URL generated for their ksvc they would modify the config-network configMap's domainTemplate value. However, with the operator any updates made directly to that config-map would eventually be overwritten by the operator's reconciliation loop.

This means that we have two different sources of truth based on how Knative was installed. This also means that the user needs to know which install mechanism was used so they can modify the correct data to get the results they want. This is not a good user experience for people manually managing Knative, nor is it good for tool authors who will want their tools to work against all Knative installs easily (and with minimal conditional logic).

It would seem to me that either:
1 - the operator should become a required component of KnServing and THE ONLY WAY to install it, and thus all configurations must be done via the CR. This would then put the burden on the operator to ensure that ALL possible configuration knobs are available via the CR since the low-level kube resources would then be off-limits to manual/tooling edits. And our docs should only point to the operator for configuring things.
2 - treat the operator as an optional tool. This would then mean it would need to probably support some kind of "policy" for how to deal with each resource in KnServing it manages. For example, within IBM we've had an operator-like system for a while and we allow each resource to have a policy that indicates whether the operator should a) just make sure the resource exists, and as long as it does then the operator won't touch it or override any edits made by the user,  or b) make sure the resource looks exactly like the operator expects and any edits made by the user will be overwritten.

I think either way can work, we just need to pick a direction. But the current state doesn't seem ideal.

Thoughts?

/cc @houshengbo @mattmoor

markusthoemmes:
I tend to disagree with this being a black and white decision.

If Knative Serving was installed without the operator, changes should definitely be made through ConfigMap edits. If Knative Serving was installed with the operator, changes should be made through the operator's API surface which is the KnativeServing CR currently.

Edits to the configuration are in the operator's persona, not the developers. I feel like the operator should know how Knative Serving was installed. The rest, I feel, is documentation of the specific platform.

duglin:
[...]

PR #165, opened 2019-09-11
Introduce immutable transforms and consider KnativeServing CR config to be canonical

### LINKED ISSUE #138: KnativeServing options are not removed when KnativeServing CR changes
Applying a KnativeServing with say a spec of

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
spec:
  config:
    logging:
      loglevel.controller: "debug"
```

And then applying the spec of:

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
```

The config is not reverted. This is the same for the other config options.

I believe it is because the way the mutations happen is that they mutate the original manifest: https://github.com/knative/serving-operator/tree/e0413caae7af488613c242dda73dfd93ee4a2b97/pkg/reconciler/knativeserving/common. They need to not mutate the original manifest and instead return a new manifest.

jcrossley3:
Fixes #138 and relates to #163 

## Proposed Changes

* Add an e2e test that reveals the problem described in #138 
* Pull in a pre-release of manifestival to transform manifests immutably

### DIFF
--- a/test/crd.go
+++ b/test/crd.go
@@ -20,5 +20,4 @@ package test
 type ResourceNames struct {
 	KnativeServing string
 	Namespace      string
-	LoggingConfig  string
 }
--- a/test/e2e/knativeservingdeployment_test.go
+++ b/test/e2e/knativeservingdeployment_test.go
@@ -17,6 +17,7 @@ package e2e
 
 import (
 	"errors"
+	"fmt"
 	"path/filepath"
 	"runtime"
 	"testing"
@@ -41,7 +42,6 @@ func TestKnativeServingDeployment(t *testing.T) {
 	names := test.ResourceNames{
 		KnativeServing: test.ServingOperatorName,
 		Namespace:      test.ServingOperatorNamespace,
-		LoggingConfig:  test.LoggingConfigMapName,
 	}
 
 	test.CleanupOnInterrupt(func() { test.TearDown(clients, names) })
@@ -87,12 +87,15 @@ func knativeServingVerify(t *testing.T, clients *test.Clients, names test.Resour
 
 // knativeServingConfigure verifies that KnativeServing config is set properly
 func knativeServingConfigure(t *testing.T, clients *test.Clients, names test.ResourceNames) {
+	// We'll arbitrarily choose the logging config
+	configKey := "logging"
+	configMapName := fmt.Sprintf("%s/config-%s", names.Namespace, configKey)
 	// Get the existing KS without any spec
 	ks, err := clients.KnativeServing().Get(names.KnativeServing, metav1.GetOptions{})
 	// Add config to its spec
 	ks.Spec = v1alpha1.KnativeServingSpec{
 		Config: map[string]map[string]string{
-			"logging": map[string]string{
+			configKey: map[string]string{
 				"loglevel.controller": "debug",
 			},
 		},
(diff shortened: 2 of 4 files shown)


garron:
Thanks @jcrossley3. I also went through this exercise yesterday and put together a similar solution. Given that your review is up, let's run with it. I'm still going to share [my branch](https://github.com/garron/serving-operator/commit/bd292c7c0f5a1f72b255999e5f2da1bdd4231f94) so that you can take a look at some slightly different options. Namely:

- It feels like ideally, the Manifestival interface wouldn't reference the Manifest type directly, and Transform would return a Manifestival. It does create some ugliness around getting the Resources because they aren't part of the interface, so I tried to keep the type assertions in code that was type aware. I'm new to go, so I could totally be off in my approach here.

- I think it might be nice to apply the transforms early in the reconciliation and then make those transformed resources available down the line (e.g. for ensuring deployments are ready). In absence of using the transformed config there, it seems easy to introduce bugs vie new code that assumes it is working with the desired manifest in that function.

I will take a more thorough look through your review to better understand some of the additional changes you made.

vendor/github.com/jcrossley3/manifestival/manifestival.go
```diff
@@ -161,14 +161,18 @@ func (f *Manifest) ResourceInterface(spec *unstructured.Unstructured) (dynamic.R
 	return f.client.Resource(mapping.Resource).Namespace(spec.GetNamespace()), nil
 }
 
-// We need to preserve the top-level target keys, specifically
-// 'metadata.resourceVersion', 'spec.clusterIP', and any existing
-// entries in a ConfigMap's 'data' field. So we only overwrite fields
-// set in our src resource.
+// We need to preserve some target keys, specifically
```
garron: I have a few thoughts/questions here. Maybe it makes sense to split this off into another review so that we can land the state management pieces.

1. This code is used by anyone who uses this library to apply a manifest. I don't know if there are other consumers of manifestival right now, but it feels a bit unfortunate that implementation decisions about how the operator should work result in functional changes to how this library interprets requests to apply a manifest. Are we sure others are OK with these?

2. As someone new to manifestival, I was surprised to see this logic and had assumed Apply() behaved similarly to "kubectl apply". Would it make sense to move this to the transform stage? That would make it possible for this library to provide a an opt-in implementation, but would also allow for altering that behavior should it not be right for a particular project. If you think what is here is universally desired behavior, I think it would be worth documenting what the intended behavior is.

3. While the comments mention a few specific examples where this is necessary, this implementation is generic and applies to all config. Regardless of whether we move it to a transform or not, it's hard for me to conclude from looking at this that the operator's config is going to act as the source of truth once run through this function. Would it be possible/practical to move to a hard-coded list of keys that we want to maintain? That would also serve as a means to ensure we're not breaking important key preservation over time. I'm thinking something like (rough first-pass idea):

```go
    my_manifest.Transform(
        KindFilter("ConfigMap", MaintainExistingValue(my_manifest, ["spec.key1", "spec.key2"]))
    ).ApplyAll()
```

jcrossley3: This code has evolved significantly from the PoC it originally was, as has my understanding of operators in general and kubernetes itself. :smile: 

You're right to assume `Apply()` behaves similarly to `kubectl apply`. It should. I would love for it to. I'm waiting for some expert to come along and tell me, "That's dumb, do it like this instead!" Are you that expert? Please!!! :smile: 

Some of the code in here came from the idea that we should allow external edits to the resources installed by manifestival. I never wanted the operator to overwrite a human's changes. I've come around in my thinking on this. And it's very possible that having the operator be the "source of truth" will possibly simplify this code. That'd be swell.

Right now, to my knowledge, only the serving-operator is using the `client-go` branch of manifestival. I know of a few other operators based on `operator-sdk/controller-runtime` that use the `master` branch without a problem, so I'm sensitive to releasing breaking changes, but not opposed to it as long as we're clear about what changes and why.

Personally, I hate the `UpdateChanged` function -- it's gone through quite a few iterations and has some fairly comprehensive tests around it because of its brittleness. I'm always open to improving it.

garron: Unfortunately, I am far from your expert on `kubectl apply`. From looking through more of the apply docs, I do see that it has it's own merging, so this may be closer to the `kubectl apply`  behavior than I had realized.

Regardless, it does seem like the merging is going to pose a challenge in making the operator's spec the source of truth. We can tackle that elsewhere though. I don't think you should block your review on this feedback.

PR #172, opened 2019-09-18
Update our dependency on K8s libs to 1.15.3

mattmoor:
With a minimum K8s version of 1.14 (starting in 0.10), 1.15.3 puts us in the center of the +/-1 version window of support.


trshafer:
Linking https://github.com/knative/serving/pull/5570

/lgtm

PR #174, opened 2019-09-19
Fix generate-yamls error by Ignoring grep error

chaodaiG:
grep returns 1 if there is no match, which is the case since there is only 1 file generated, which was negated by `-v`. Fixing it by ignoring grep error

/cc @k4leung4


chaodaiG:
@k4leung4 , an /approve is also needed

PR #178, opened 2019-09-20
Additional tests for resource deletion and update

mgencur:
Fixes #

## Proposed Changes

* Instead of checking that just deployments are removed, check that all resources deployed by the operator are removed when KnativeServing CR is removed
* Add more tests for updating KnativeServing and propagating the changes to the target config map:
    - delete a single key/value pair
    - use an empty map as the value

**Release Note**

```release-note

```


mgencur:
@jcrossley3 thanks for all the reviews;-)

PR #180, opened 2019-09-25
Set the default GroupVersionKind for KnativeServing

houshengbo:
Fixes #

## Proposed Changes
* The CR KnativeServing can have a default GroupVersionKind, so that we do not need to set it in the reconcile function.

**Release Note**

```release-note

```


houshengbo:
@k4leung4 We have not got any unit test ready, but this is a good start to add unit tests.

PR #185, opened 2019-10-08
[WIP] Factor platform-specific code out into its own operator

### LINKED ISSUE #122: Platform specific configuration for operator
Each platform has some amount of customization required with their installation.
It is unclear what approach is ideal to support these types of customization.

@houshengbo kindly documented some of the thoughts on this topic in this doc.

https://docs.google.com/document/d/1OtlkJ8whTgHYTz5rz0lq5mNdDGK_pakG/edit

jcrossley3:
Fixes #122 

## Proposed Changes

* Delete platform-specific transformation code
* Add a `Deployment` for a platform-specific operator that fires up a webhook to tailor the `KnativeServing` CR prior to consumption by the serving-operator

The minikube image fires up a webhook if it detects it's running in
minikube -- it does nothing otherwise -- and sets
`istio.sidecar.includeOutboundIPRanges` to "10.0.0.1/24" if it's not
already set. Its source repo is https://github.com/jcrossley3/k-s-o-minikube

The OpenShift image configures ingress, egress, and the revision-url-template, as well as a validation webhook for namespace and minimum version. Its source repo is https://github.com/openshift-knative/knative-serving-openshift

Neither operator will do anything if it doesn't recognize its platform.


garron:
@jcrossley3 Is there a design document which goes over your approach here?

After my initial look through this, the primary question in my mind is: If we're going to need to build a custom operator for every platform, do we really need to have two operators (non-platform-specific and platform-specific) running and acting on these resources at the same time? Or would it be possible to just run a single platform-specific operator? (e.g. by refactoring the existing operator into a library intended for implementing platform-specific operators)

PR #186, opened 2019-10-09
Move the CRD to the directory config

### LINKED ISSUE #187: The creation of CRD should be made part of the operator installation
We currently install CRD as prerequisite of the operator installation. This induced that the CRD yaml is not included in the operator release yaml. We should make sure the operator verifies the existence of the CRD and install the CRD automatically when we install the operator.

houshengbo:
Fixes #187

## Proposed Changes

This PR makes the CRD installation a part of the operator installation, instead of
a prerequisite. The operator installs the CRD automatically.

**Release Note**

```release-note

```


houshengbo:
@jcrossley3 We need to move the CRD yaml under config anyway, to make sure it will be included in the release yaml, generated by ko commands.

PR #191, opened 2019-10-23
Add a new script to test the latest generated manifest of knative serving

### LINKED ISSUE #188: Check whether the operator works correctly with the latest knative serving yaml
We can either create a PR-based job or periodic job to generate the latest release yaml for knative serving from master, and run the integration tests, to see if the operator works correctly with the latest knative serving.

This can facilitate the future operator release as well, since we can track the knative serving yaml on a timely fashion.

houshengbo:
Fixes #188 

## Proposed Changes

* Added a new script called e2e-tests-latest-serving.sh, to run the operator installation with the latest generated manifest of knative serving, and run the integration tests against the Kube cluster. This script will be called directly by a periodic prow job.

**Release Note**

```release-note

```

### DIFF
--- a/test/e2e-tests-latest-serving.sh
+++ b/test/e2e-tests-latest-serving.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script runs the end-to-end tests against Knative Serving
+# Operator built from source. However, this script will download the latest
+# source code of knative serving and generated the latest manifest file for
+# serving installation. The serving operator will use this newly generated
+# manifest file to replace the one under the directory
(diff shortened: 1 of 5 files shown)


test/e2e-tests-latest-serving.sh
```diff
@@ -0,0 +1,85 @@
...
+  # Go the directory to download the source code of knative serving
+  cd $KNATIVE_SERVING_DIR
+
+  # Download the source code of knative serving
+  git clone https://github.com/knative/serving.git
+  cd serving
+  mkdir -p output
+
+  # Generate the manifest
+  export YAML_OUTPUT_DIR=$KNATIVE_SERVING_DIR/serving/output
```
k4leung4: why does this need to be exported?

test/e2e-tests-latest-serving.sh
```diff
@@ -0,0 +1,85 @@
...
+# the cluster and run the tests.
+
+# Calling this script without arguments will create a new cluster in
+# project $PROJECT_ID, start knative in it, run the tests and delete the
+# cluster.
+
+source $(dirname $0)/e2e-common.sh
+
+OPERATOR_DIR=$(dirname $0)/..
+KNATIVE_SERVING_DIR=$OPERATOR_DIR/..
```
k4leung4: can you enclose the variable with { }, as in 
KNATIVE_SERVING_DIR=${OPERATOR_DIR}/..

This will make it easier to read what the var is, same for the rest of the file.

test/e2e-tests-latest-serving.sh
houshengbo: We used the script called generate-yamls.yaml in knative serving, which expected an env var YAML_OUTPUT_DIR as where we generate the yamls, so I export it, making sure the script will know it exists. Otherwise, a random dir will be generated and used.

test/e2e-tests-latest-serving.sh
houshengbo: OK

PR #198, opened 2019-10-29
Move to operator.knative.dev API_Group

### LINKED ISSUE #197: Inconsisten API group
see discussions here: https://github.com/knative/eventing-operator/issues/18

## API Group

I'd propose that we change the `apiGroup` of the serving operator from `serving.knative.dev` to `operators.knative.dev`.

Motivation:
* cleaner separation of APIs (operator API vs core serving APIs)
* simplify RBAC when operator CRDs and serving CRDs are not in the same group
* eventing operator uses `operator.knative.dev`

## KIND

the current `kind` is `KnativeServing` ... I think this makes sense... more sense than _just_ `Serving`, but IMO we should align w/ the `eventing-operator` ... which currently names it `Eventing`... perhaps that should be changed to `KnativeEventing` (as it was before) .

matzew:
Fixes #197

## Proposed Changes

* move to `operator.knative.dev` Api Group

**Release Note**

```release-note
KnativeServing operator is moved to a new operator.knative.dev API Group
```


houshengbo:
@matzew Thx for the PR.


 knative.dev/serving-operator/pkg/reconciler/knativeserving
# knative.dev/serving-operator/pkg/reconciler/knativeserving
pkg/reconciler/knativeserving/knativeserving_controller.go:147:47: r.Base.KnativeServingClientSet.ServingV1alpha1 undefined (type "knative.dev/serving-operator/pkg/client/clientset/versioned".Interface has no field or method ServingV1alpha1)
# knative.dev/serving-operator/test
test/clients.go:33:13: undefined: "knative.dev/serving-operator/pkg/client/clientset/versioned/typed/serving/v1alpha1".ServingV1alpha1Interface 

I think at least you missed these two places.

bbrowning:
How do we propose to handle upgrades from our first operator release to this one with this change?

PR #204, opened 2019-11-01
Add the marker of the latest commit ID for better debugging

houshengbo:
## Proposed Changes

* This PR will help us find the commit ID directly, if the periodic prow job fails on a certain commit of serving.

**Release Note**

```release-note

```


n3wscott:
nice work 👍

PR #211, opened 2019-11-08
Clean up the [un]structured conversions a bit

jcrossley3:
I think this is "The Right Way" to do it, but I could be wrong. Makes
the code a bit smaller anyway.

Fixes https://github.com/knative/serving-operator/pull/20#discussion_r297817556


trshafer:
Thanks for doing this!

/lgtm
/approve

PR #212, opened 2019-11-12
We need to fail the install if any transforms error

jcrossley3:
Otherwise, updating an existing `KnativeServing` resource in a way that makes any of the transformations fail will cause the operator to retry indefinitely, but the status of the `KnativeServing` CR will continue to show `Ready`.

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -120,6 +120,7 @@ func (r *Reconciler) reconcile(ctx context.Context, ks *servingv1alpha1.KnativeS
 
 	manifest, err := r.transform(ks)
 	if err != nil {
+		ks.Status.MarkInstallFailed(err.Error())
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -181,7 +182,6 @@ func (r *Reconciler) install(manifest *mf.Manifest, instance *servingv1alpha1.Kn
 // Check for all deployments available
 func (r *Reconciler) checkDeployments(manifest *mf.Manifest, instance *servingv1alpha1.KnativeServing) error {
 	r.Logger.Debug("Checking deployments")
-	defer r.updateStatus(instance)
```
houshengbo: @jcrossley3 Has this line intentionally been removed?

jcrossley3: I think it's redundant due to https://github.com/knative/serving-operator/blob/39757e9cc8d2d0c94096180b1ee886e84d5a4455/pkg/reconciler/knativeserving/knativeserving_controller.go#L97

Issue #5, opened 2019-06-13
The serving-operator should export prometheus metrics


trshafer:
This will require a creating a `Service` that exposes a metrics port similar to the knative serving controller (https://github.com/knative/serving/blob/9bcd6f482377c5c42665e7d5fad9072a180301fc/config/400-controller-service.yaml#L25)

Example service:
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app: serving-operator
    serving.knative.dev/release: devel
  name: serving-operator
  namespace: default
spec:
  ports:
  - name: metrics
    port: 9090
    protocol: TCP
    targetPort: 9090
  selector:
    app: serving-operator
```

jihuin:
Hey, the operator now can export either Prometheus metrics or Stackdriver metrics based on configuration. Would you mind closing this issue?

PR #216, opened 2019-11-13
Add metrics port to operator

### LINKED ISSUE #192: Operator should emit metrics
I don't believe the operator has metrics that it emits around installing knative serving or receiving a new CRD.

jihuin:
Fixes #192, #5 

## Proposed Changes

* Added metrics port in `operator.yaml` so the operator can report metrics to 9090 port.
* Changed the component name from `serving-operator` to `serving_operator` because `-` is not allowed in metric name for prometheus exporter.

Currently the operator only emits the metrics on **reconcile count**, **reconcile latency** and **work queue depth**, that are inherited from `knative.dev/controller`.

**Release Note**

```release-note
NONE
```


cmd/manager/main.go
```diff
@@ -32,5 +32,5 @@ func main() {
 	if err != nil {
 		log.Fatal("Error building kubeconfig", err)
 	}
-	sharedmain.MainWithConfig(signals.NewContext(), "serving-operator", cfg, knativeserving.NewController)
+	sharedmain.MainWithConfig(signals.NewContext(), "operator", cfg, knativeserving.NewController)
```
houshengbo: Why do you change the name here?

config/operator.yaml
```diff
@@ -32,3 +32,10 @@ spec:
                   fieldPath: metadata.namespace
             - name: METRICS_DOMAIN
               value: knative.dev/serving-operator
+            - name: CONFIG_LOGGING_NAME
+              value: config-logging
+            - name: CONFIG_OBSERVABILITY_NAME
+              value: config-observability
```
houshengbo: I do not think we need to add CONFIG_LOGGING_NAME and CONFIG_OBSERVABILITY_NAME, anymore.
By default, knative/pkg will pick config-logging and config-observability.

houshengbo:
@jihuin Is there any test we can leverage to check if metrics are exposed?
Is 9090 the only port designated to metrics or is it something we can configure?

cmd/manager/main.go
jihuin: Because the component name `serving-operator` will be passed to the prometheus exporter as a prefix in the metric name, but the symbol `-` is not allowed in the prometheus metric name. Meanwhile, single-word application/component name is recommended by prometheus and all other components in knative serving have single-word names.

config/operator.yaml
jihuin: It's true that knative/pkg will watch config-logging and config-observability configmaps by default if nothing is specified. Maybe it is meaningful to add here to let the users be aware of the purposes of these two configmaps.

jihuin: Though the configmap filenames are quite self-explaining, the users can not see if or where the configmaps are watched unless digging into knative/pkg

jihuin:
@houshengbo The metrics are reported through `knative.dev/pkg/controller`, and are tested in the same package: https://github.com/knative/pkg/blob/master/controller/stats_reporter_test.go

The port number is configurable, and 9090 is the default port number of Prometheus server.

cmd/manager/main.go
houshengbo: Is serving.operator supported? I lean on to keep the full name.

config/operator.yaml
houshengbo: https://github.com/knative/pkg/commit/94911510c3dc726029d2a1c0af8a5e31301eb2d7 
I just had this PR merged. If config-logging and config-observability are missing, the controller should be launched without problem. With this PR, would it be fine, if  config-logging and config-observability are missing??

houshengbo:
@jihuin I would like to test this change locally, could you tell me the steps? Thx.

jihuin:
@houshengbo Steps to test metrics with Prometheus exporter (default):

1. Forward the local port 9090 to the port 9090 on operator pod via the command
`kubectl port-forward ${{operator_pod_name}} 9090`

2. Check the metrics by visiting `localhost:9090/metrics` in the browser, or run `curl localhost:9090/metrics` in the terminal

cmd/manager/main.go
jihuin: Only letters, digits and underscores are supported, and single-word prefix is recommended. See  https://prometheus.io/docs/practices/naming/

Since all knative serving components have single-word names, e.g. webhook, activator, autoscalar, it makes sense to align with the current naming pattern.

matzew: Good catch @jihuin I noticed the same, and I know that **ALL** other knative components _NOW_  using `_` underscore.  For instance:
* https://github.com/knative/eventing/commit/36058f4fe735902265b63ab43c3f82ff289d07ef

Since we have more than one operator, let's please use `serving_operator` !

config/operator.yaml
matzew: @jihuin again, you are correct here. 

See https://github.com/knative/eventing/commit/36058f4fe735902265b63ab43c3f82ff289d07ef for adding port (the `CONFIG_XYZ_NAME` were already in the yamls, eg:

https://github.com/knative/eventing/blob/master/config/channels/in-memory-channel/500-controller.yaml#L37-L40

matzew:
@jihuin can u report on your CLA status ?

jihuin:
@googlebot rescan

cmd/manager/main.go
houshengbo: We can use the underscore.

matzew: @jihuin can you sent a PR to use `serving_operator` with underscore ?

config/operator.yaml
jihuin: If the configmaps are missing and not watched, the exporter won't be set up and no metric is exported, but no information is provided in the logs. The configs here can be informative.

cmd/manager/main.go
jihuin: Changed to `serving_operator`

jihuin:
fix #5

matzew:
@n3wscott I think this now looks good... mind approving ? 

I am not an approver here, on the serving operator :crying_cat_face:

PR #223, opened 2019-11-14
Letting flags be changed from outside 

cardil:
## Proposed Changes

* *Letting flags be changed from outside*
  Changing those flags is needed to perform tests on RH operators.
* *Adding Intellij, and vandors excluded files to gitignore*

### Implementation

This PR let change test flags from outside. It can be done either by:

 * declaring environmental variables: `TEST_NAMESPACE`, and `TEST_RESOURCE`,
 * using `-ldflags "-X knative.dev/serving-operator/test.ServingOperatorNamespace=knative-serving"`

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Temporary Build Files
 build/_output
 build/_test
+# Goland
+.idea
+
 # Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Emacs ###
 # -*- mode: gitignore; -*-
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -19,9 +19,19 @@ limitations under the License.
 
 package test
 
-const (
+import "os"
+
+var (
 	// ServingOperatorNamespace is the default namespace for serving operator e2e tests
-	ServingOperatorNamespace = "operator-tests"
+	ServingOperatorNamespace = getenv("TEST_NAMESPACE", "operator-tests")
 	// ServingOperatorName is the default operator name for serving operator e2e tests
-	ServingOperatorName = "knative-serving"
+	ServingOperatorName = getenv("TEST_RESOURCE", "knative-serving")
 )
+
+func getenv(name, defaultValue string) string {
+	value, set := os.LookupEnv(name)
+	if !set {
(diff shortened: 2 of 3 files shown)


cardil:
This PR mirrors https://github.com/knative/eventing-operator/pull/21

.gitignore
```diff
@@ -1,6 +1,13 @@
 # Temporary Build Files
 build/_output
 build/_test
+# Goland
+.idea
```
matzew: not 100% sure on getting this in.... I guess... shouldn't those be on your own global `.gitignore` ?

.gitignore
```diff
@@ -1,6 +1,13 @@
...
+# Vendors OWNERS files
+vendor/**/OWNERS
+# Test Infra Dev Stats
+vendor/knative.dev/test-infra/devstats
+
```
matzew: why is that ?

test/e2e_flags.go
```diff
@@ -19,9 +19,19 @@ limitations under the License.
...
-	ServingOperatorName = "knative-serving"
+	ServingOperatorName = getenv("TEST_RESOURCE", "knative-serving")
 )
+
+func getenv(name, defaultValue string) string {
+	value, set := os.LookupEnv(name)
+	if !set {
+		value = defaultValue
+	}
+	return value
```
matzew: :+1:

.gitignore
houshengbo: I would say not change this .gitignore by far.
The script under /hack/update-deps.sh is supposed to update the dependencies, and remove the useless files, including OWNERS. However, some packages may keep them, based on the configuration in Gopkg.toml. Let's keep the dependencies(with all the files) based what is generated by /hack/update-deps.sh.

.gitignore
cardil: You mean by using `git config --global core.excludesfile` option? 

I think not many people actually use that. I didn't hear of that option, nor any of developers in my office (I know that's not representative :wink:)

On top that it is easier for newcomers, so that repo declares those ignores.

.gitignore
cardil: Removing those.

PR #232, opened 2019-11-21
Added reconciler test and updated deps

jihuin:
## Proposed Changes

* Added reconciler test based on https://github.com/knative/serving/blob/master/pkg/reconciler/reconciler_test.go
* Updated deps by `dep ensure` and `./hack/update-deps.sh`

**Release Note**

```release-note
NONE
```


pkg/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+package reconciler
+
+import (
+	"context"
+	"testing"
+
+	_ "knative.dev/pkg/client/injection/client/fake"
+    _ "knative.dev/pkg/client/injection/kube/client/fake"
```
trshafer: nit: spacing

jihuin: resolved

anniefu:
It might be worth minimizing the change to only the package needed for this test instead of updating all of them. i.e. "dep ensure -update knative.dev/pkg"

Running ./update-deps.sh to update all of vendor/ should probably be isolated to its own PR to keep updating misc. packages and adding new feature/test code in independent commits that can be reverted without impact each other.

jihuin:
@anniefu  The changes are all about adding fake clients that are required for the test.

PR #233, opened 2019-11-22
Add metrics about KnativeServing CR edits

jihuin:
## Proposed Changes

Serving operator can emit metrics on KnativeServing CR edits, including creation, modification and deletion.

Steps to test metrics with Prometheus exporter (default):

1. Forward the local port 9090 to the port 9090 on operator pod by running
`kubectl port-forward ${{operator_pod_name}} 9090`
2. Check the metrics by visiting `localhost:9090/metrics`

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -28,7 +28,7 @@ import (
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/sets"
+
 	"k8s.io/client-go/tools/cache"
 
 	"knative.dev/pkg/controller"
@@ -40,7 +40,10 @@ import (
 )
 
 const (
-	finalizerName = "delete-knative-serving-manifest"
+	finalizerName  = "delete-knative-serving-manifest"
+	creationChange = "creation"
+	editChange     = "edit"
+	deletionChange = "deletion"
 )
 
 var (
@@ -54,7 +57,7 @@ type Reconciler struct {
 	// Listers index properties about resources
 	knativeServingLister listers.KnativeServingLister
 	config               mf.Manifest
-	servings             sets.String
+	servings             map[string]int64
 }
 
 // Check that our Reconciler implements controller.Reconciler
@@ -74,17 +77,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
 	original, err := r.knativeServingLister.KnativeServings(namespace).Get(name)
 	if apierrs.IsNotFound(err) {
 		return nil
-
 	} else if err != nil {
 		r.Logger.Error(err, "Error getting KnativeServing")
 		return err
 	}
 	if original.GetDeletionTimestamp() != nil {
-		r.servings.Delete(key)
+		if _, ok := r.servings[key]; ok {
+			delete(r.servings, key)
+			r.StatsReporter.ReportKnativeServingChange(key, deletionChange)
+		}
 		return r.delete(original)
 	}
-	// Keep track of the number of KnativeServings in the cluster
-	r.servings.Insert(key)
+	// Keep track of the number and generation of KnativeServings in the cluster.
+	newGen := original.Generation
+	oldGen, ok := r.servings[key]
+	if !ok && newGen == 1 {
+		r.StatsReporter.ReportKnativeServingChange(key, creationChange)
+	} else if ok && newGen == oldGen+1 {
+		r.StatsReporter.ReportKnativeServingChange(key, editChange)
+	}
+	r.servings[key] = newGen
 
 	// Don't modify the informers copy.
 	knativeServing := original.DeepCopy()
@@ -231,7 +243,7 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
 	if len(instance.GetFinalizers()) == 0 || instance.GetFinalizers()[0] != finalizerName {
 		return nil
 	}
-	if r.servings.Len() == 0 {
+	if len(r.servings) == 0 {
 		if err := r.config.DeleteAll(&metav1.DeleteOptions{}); err != nil {
 			return err
 		}
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package knativeserving
 
 import (
+	"fmt"
 	"context"
 
 	mf "github.com/jcrossley3/manifestival"
@@ -84,17 +85,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
 	if original.GetDeletionTimestamp() != nil {
 		if _, ok := r.servings[key]; ok {
 			delete(r.servings, key)
-			r.StatsReporter.ReportKnativeServingChange(key, deletionChange)
+			r.StatsReporter.ReportKnativeservingChange(key, deletionChange)
 		}
 		return r.delete(original)
 	}
 	// Keep track of the number and generation of KnativeServings in the cluster.
 	newGen := original.Generation
-	oldGen, ok := r.servings[key]
-	if !ok && newGen == 1 {
-		r.StatsReporter.ReportKnativeServingChange(key, creationChange)
-	} else if ok && newGen == oldGen+1 {
-		r.StatsReporter.ReportKnativeServingChange(key, editChange)
+	if oldGen, ok := r.servings[key]; ok {
+		if newGen > oldGen {
+			r.StatsReporter.ReportKnativeservingChange(key, editChange)
+		} else if newGen < oldGen {
+			return fmt.Errorf("reconciling obsolete generation of KnativeServing %s: newGen = %d and oldGen = %d", key, newGen, oldGen)
+		}
+	} else {
+		// No metrics are emitted when newGen > 1: the first reconciling of
+		// a new operator on an existing KnativeServing resource.
+		if newGen == 1 {
+			r.StatsReporter.ReportKnativeservingChange(key, creationChange)
+		}
 	}
 	r.servings[key] = newGen
 
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -104,7 +104,6 @@ func NewBase(ctx context.Context, controllerAgentName string, cmw configmap.Watc
 	}
 
 	// Create metrics reporter
-	logger.Debug("Creating stats reporter")
 	statsReporter, err := NewStatsReporter(controllerAgentName)
 	if err != nil {
 		logger.Fatal(err)
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -62,6 +62,9 @@ type Base struct {
 	// Kubernetes API.
 	Recorder record.EventRecorder
 
+	// StatsReporter reports reconciler's metrics.
+	StatsReporter StatsReporter
+
 	// Sugared logger is easier to use but is not as performant as the
 	// raw logger. In performance critical paths, call logger.Desugar()
 	// and use the returned raw logger instead. In addition to the
@@ -100,6 +103,13 @@ func NewBase(ctx context.Context, controllerAgentName string, cmw configmap.Watc
 		}()
 	}
 
+	// Create metrics reporter
+	logger.Debug("Creating stats reporter")
+	statsReporter, err := NewStatsReporter(controllerAgentName)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
 	base := &Base{
 		KubeClientSet:           kubeClient,
 		SharedClientSet:         sharedclient.Get(ctx),
@@ -107,6 +117,7 @@ func NewBase(ctx context.Context, controllerAgentName string, cmw configmap.Watc
 		DynamicClientSet:        dynamicclient.Get(ctx),
 		ConfigMapWatcher:        cmw,
 		Recorder:                recorder,
+		StatsReporter:           statsReporter,
 		Logger:                  logger,
 	}
 
--- a/pkg/reconciler/stats_reporter.go
+++ b/pkg/reconciler/stats_reporter.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconciler
+
+import (
+	"context"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"knative.dev/pkg/metrics"
+)
+
+const (
+	knativeServingChangeCountName = "knativeserving_change_count"
+)
+
+var (
+	knativeServingChangeCountStat = stats.Int64(
+		knativeServingChangeCountName,
+		"Number of KnativeServing changes including creation, edit and deletion",
+		stats.UnitDimensionless)
+	// Create the tag keys that will be used to add tags to our measurements.
+	// Tag keys must conform to the restrictions described in
+	// go.opencensus.io/tag/validate.go. Currently those restrictions are:
+	// - length between 1 and 255 inclusive
+	// - characters are printable US-ASCII
+	reconcilerTagKey     = tag.MustNewKey("reconciler")
+	knativeServingTagKey = tag.MustNewKey("knativeServing")
+	changeTagKey         = tag.MustNewKey("change")
+)
+
+func init() {
+	// Create views to see our measurements. This can return an error if
+	// a previously-registered view has the same name with a different value.
+	// View name defaults to the measure name if unspecified.
+	views := []*view.View{{
+		Description: knativeServingChangeCountStat.Description(),
+		Measure:     knativeServingChangeCountStat,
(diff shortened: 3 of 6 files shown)


pkg/reconciler/stats_reporter_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	countWas := int64(0)
+	if m := getMetric(t, InstallSuccessCountName); m != nil {
+		countWas = m.Data.(*view.CountData).Value
+	}
+	expectedTags := map[string]string{
+		keyTagKey.Name():        fmt.Sprintf("%s/%s", testResourceNamespace, testResourceName),
+		reconcilerTagKey.Name(): reconcilerMockName,
+	}
+
```
mattmoor: Remove trailing whitespace:
```suggestion

```

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &reporter{reconciler: reconciler, ctx: ctx}, nil
+}
+
+// ReportInstallSuccess reports
+func (r *reporter) ReportInstallSuccess(resourceNamespace, resourceName string, duration time.Duration) error {
+	key := fmt.Sprintf("%s/%s", resourceNamespace, resourceName)
```
trshafer: generating the context should be a separate function that is shared with other metrics calls.

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,172 @@
...
+
+// WithStatsReporter attaches the given StatsReporter to the provided context
+// in the returned context.
+func WithStatsReporter(ctx context.Context, sr StatsReporter) context.Context {
+	return context.WithValue(ctx, srKey{}, sr)
+}
+
+// GetStatsReporter attempts to look up the StatsReporter on a given context.
+// It may return null if none is found.
+func GetStatsReporter(ctx context.Context) StatsReporter {
```
anniefu: Is it necessary to associate StatsReporters with contexts? I believe contexts are intended for cross-process usage (https://golang.org/pkg/context/), but StatsReporters will only be used within the one reconciler controller.

pkg/reconciler/reconciler.go
```diff
@@ -100,13 +103,24 @@ func NewBase(ctx context.Context, controllerAgentName string, cmw configmap.Watc
...
+	if statsReporter == nil {
+		logger.Debug("Creating stats reporter")
+		var err error
+		statsReporter, err = NewStatsReporter(controllerAgentName)
+		if err != nil {
+			logger.Fatal(err)
+		}
+	}
+
 	base := &Base{
```
anniefu: Could you try switching this to use NewBase constructor function, and move initializing the StatsReporter to that function.

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,172 @@
...
+
+const (
+	InstallSuccessCountName    = "install_success_count"
+	InstallSuccessLatencyName  = "install_success_latency"
+	DeploymentReadyCountName   = "deployment_ready_count"
+	DeploymentReadyLatencyName = "deployment_ready_latency"
+)
+
+var (
+	InstallSuccessCountStat = stats.Int64(
```
anniefu: Why only track success? Failure to install could be a useful metric too.

Maybe make the metric "InstallCount" and then tag the metric with a "success" label that can be true/false? That would allow getting total #install attempts, then filtering by tag to get #successes & #failures.

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -206,6 +211,10 @@ func (r *Reconciler) checkDeployments(manifest *mf.Manifest, instance *servingv1
 			}
 		}
 	}
+	if instance.Status.IsDeploying() {
```
anniefu: Mmh, what does IsDeploying actually check?

To me this if block reads as "if still deploying, report deployment is read", which seems preemptive. It's possible that IsDeploying()'s implementation is really IsDeployed() though.

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,172 @@
...
+			TagKeys:     []tag.Key{reconcilerTagKey, keyTagKey},
+		},
+	); err != nil {
+		panic(err)
+	}
+}
+
+// StatsReporter defines the interface for sending metrics
+type StatsReporter interface {
+	// ReportKnativeServingReady reports the count and latency metrics for a reconcile operation
```
anniefu: These metrics seem like they are redundant. Should an install be considered successful before the deployments are ready?

pkg/reconciler/stats_reporter.go
jihuin: Refactored

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,172 @@
...
+			Aggregation: view.Count(),
+			TagKeys:     []tag.Key{reconcilerTagKey, keyTagKey},
+		},
+		&view.View{
+			Description: DeploymentReadyLatencyStat.Description(),
+			Measure:     DeploymentReadyLatencyStat,
+			Aggregation: view.Distribution(10, 100, 1000, 10000, 30000, 60000), // Bucket boundaries are 10ms, 100ms, 1s, 10s, 30s and 60s.
+			TagKeys:     []tag.Key{reconcilerTagKey, keyTagKey},
+		},
+	); err != nil {
```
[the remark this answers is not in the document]
jihuin: Separated the views creation and registration processes

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -174,6 +175,10 @@ func (r *Reconciler) install(manifest *mf.Manifest, instance *servingv1alpha1.Kn
 		instance.Status.MarkInstallFailed(err.Error())
 		return err
 	}
+	if !instance.Status.IsInstalled() {
+		duration := time.Since(instance.Status.LastInstallStatusTransitionTime())
```
anniefu: Maybe adding a timer to the if statement above is the best way to derive duration.

If the last three installs were  T1: Success, T2: Fail, T3: Fail
Then the duration here for T3 would represent the time between T3 and T1, right?

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,172 @@
...
+	InstallSuccessCountName    = "install_success_count"
+	InstallSuccessLatencyName  = "install_success_latency"
+	DeploymentReadyCountName   = "deployment_ready_count"
+	DeploymentReadyLatencyName = "deployment_ready_latency"
+)
+
+var (
+	InstallSuccessCountStat = stats.Int64(
+		InstallSuccessCountName,
+		"Number of successful KnativeServing install",
```
anniefu: Is this what we want "install" to mean? The number of installs of the KnativeServing custom resource? Or do we want it to mean the install of knative-serving itself?

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,173 @@
...
+	"fmt"
+	"time"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"knative.dev/pkg/metrics"
+)
+
+const (
```
anniefu: Do the consts and vars need to be exported?

pkg/reconciler/knativeserving/knativeserving_controller.go
jihuin: You will execute this block only when satisfying both
1. install succeeded, otherwise the err would be returned in the previous block
2. install status was not marked success. Because in one install like T1, the status changes as unknown (initialized) -> fail (installing) -> success (installed), and I want to capture the change from fail to success.

pkg/reconciler/stats_reporter.go
jihuin: With operator, we install knative-serving by installing KnativeServing custom resource. Unless the users install knative-serving by themselves, these two numbers should be the same. And what we actually care about, in my opinion, is the usage of operator to install KnativeServing custom resource.

pkg/reconciler/knativeserving/knativeserving_controller.go
jihuin: `IsDeploying()` returns true when the manifest install succeeded but the deployment is not ready. Adding this block here is to capture the latency time before marking the deployment as ready in the next line.

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+		context.Background(),
+		tag.Insert(reconcilerTagKey, r.reconciler),
+		tag.Insert(keyTagKey, key))
+	if err != nil {
+		return nil, err
+	}
+	return ctx, nil
+}
+
+// ReportInstallSuccess captures successful KnativeServing CR install
```
trshafer: nit: comment is unclear

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+		context.Background(),
+		tag.Insert(reconcilerTagKey, reconciler))
+	if err != nil {
+		return nil, err
+	}
+	return &reporter{reconciler: reconciler, ctx: ctx}, nil
+}
+
+func(r *reporter) getContext(key string) (context.Context, error) {
+	ctx, err := tag.New(
```
trshafer: can this just return tag.New?

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -70,18 +70,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
 	original, err := r.knativeServingLister.KnativeServings(namespace).Get(name)
 	if apierrs.IsNotFound(err) {
 		// The resource was deleted
-		r.servings.Delete(key)
-		if r.servings.Len() == 0 {
-			r.config.DeleteAll(&metav1.DeleteOptions{})
+		// Report the KnativeServing edit when deleting an existed KnativeServing
+		if _, ok := r.servings[key]; ok {
+			r.StatsReporter.ReportKnativeServingEdit(key)
```
trshafer: should this be a deletion metric? and should it be tied to r.config.DeleteAll?

trshafer: test?

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -70,18 +70,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 	}
-	// Keep track of the number of KnativeServings in the cluster
-	r.servings.Insert(key)
+	// Keep track of the number and generation of KnativeServings in the cluster
+	// Report the KnativeServing edit when creating a new KnativeServing or
+	// changing an existed KnativeServing
+	newGen := original.Generation
+	oldGen, ok := r.servings[key]
+	if (ok && newGen == oldGen+1) || (!ok && newGen == 1) {
+		r.StatsReporter.ReportKnativeServingEdit(key)
```
trshafer: Can the generation be attached?

pkg/reconciler/reconciler.go
```diff
@@ -100,13 +103,21 @@ func NewBase(ctx context.Context, controllerAgentName string, cmw configmap.Watc
 		}()
 	}
 
+	// Create metrics reporter
+	logger.Debug("Creating stats reporter")
```
anniefu: This log message is probably overkill.

pkg/reconciler/reconciler.go
```diff
@@ -100,13 +103,21 @@ func NewBase(ctx context.Context, controllerAgentName string, cmw configmap.Watc
...
+	statsReporter, err := NewStatsReporter(controllerAgentName)
+	if err != nil {
+		logger.Fatal(err)
```
anniefu: Does this exit the controller? Is that the desired behavior or should we log an error and then proceed without metrics?

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+const (
+	knativeServingEditCountName = "knativeserving_edit_count"
+)
+
+var (
+	knativeServingEditCountStat = stats.Int64(
+		knativeServingEditCountName,
+		"Number of KnativeServing changes",
```
anniefu: Could you list off the type of edits being included in this count in the description? i.e. create, delete, modify, etc.

[nit] Use "edits" instead of "changes" for consistency.

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+// StatsReporter defines the interface for sending metrics
+type StatsReporter interface {
+	// ReportKnativeServingReady reports the count and latency metrics for a reconcile operation
+	ReportKnativeServingEdit(NamespacedName string) error
+}
+
+// Reporter holds cached metric objects to report metrics
```
anniefu: [nit] reporter in comment should be lower-case to match its unexported status

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+type reporter struct {
+	reconciler string
+	ctx        context.Context
+}
+
+// NewStatsReporter creates a reporter that collects and reports metrics
```
anniefu: [nit] golang style for function comments recommends periods at the end. https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+		tag.Insert(reconcilerTagKey, r.reconciler),
+		tag.Insert(keyTagKey, key))
+	if err != nil {
+		return nil, err
+	}
+	return ctx, nil
+}
+
+// ReportInstallSuccess captures successful KnativeServing CR install
+func (r *reporter) ReportKnativeServingEdit(key string) error {
```
anniefu: There should be validation of the values that are passed in here. The valid values should be defined as const strings somewhere in this file that callers can re-use.

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+
+// StatsReporter defines the interface for sending metrics
+type StatsReporter interface {
+	// ReportKnativeServingReady reports the count and latency metrics for a reconcile operation
+	ReportKnativeServingEdit(NamespacedName string) error
+}
+
+// Reporter holds cached metric objects to report metrics
+type reporter struct {
+	reconciler string
```
anniefu: [nit] reconcilerName?

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+		tag.Insert(reconcilerTagKey, reconciler))
+	if err != nil {
+		return nil, err
+	}
+	return &reporter{reconciler: reconciler, ctx: ctx}, nil
+}
+
+func(r *reporter) getContext(key string) (context.Context, error) {
+	ctx, err := tag.New(
+		context.Background(),
```
anniefu: Shouldn't this be using the reporter struct's context?

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,106 @@
...
+		knativeServingEditCountName,
+		"Number of KnativeServing changes",
+		stats.UnitDimensionless)
+	// Create the tag keys that will be used to add tags to our measurements.
+	// Tag keys must conform to the restrictions described in
+	// go.opencensus.io/tag/validate.go. Currently those restrictions are:
+	// - length between 1 and 255 inclusive
+	// - characters are printable US-ASCII
+	reconcilerTagKey = tag.MustNewKey("reconciler")
+	keyTagKey        = tag.MustNewKey("key")
```
anniefu: Is this intended to represent the type of edit being made? If so, maybe "editType" is the better tag key?

pkg/reconciler/stats_reporter_test.go
```diff
@@ -0,0 +1,82 @@
...
+	if err = reporter.ReportKnativeServingEdit(key); err != nil {
+		t.Error(err)
+	}
+	metricstest.CheckCountData(t, knativeServingEditCountName, expectedTags, countWas+1)
+}
+
+func getMetric(t *testing.T, metric string) *view.Row {
```
anniefu: This function can be condensed to (from https://github.com/knative/pkg/blob/master/controller/stats_reporter_test.go)
	initialReconcileCount := int64(0)
	if d, err := view.RetrieveData("reconcile_count"); err == nil && len(d) == 1 {
		initialReconcileCount = d[0].Data.(*view.CountData).Value
	}

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -74,17 +77,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 		return r.delete(original)
 	}
-	// Keep track of the number of KnativeServings in the cluster
-	r.servings.Insert(key)
+	// Keep track of the number and generation of KnativeServings in the cluster.
+	newGen := original.Generation
+	oldGen, ok := r.servings[key]
+	if !ok && newGen == 1 {
+		r.StatsReporter.ReportKnativeServingChange(key, creationChange)
+	} else if ok && newGen == oldGen+1 {
```
anniefu: What about the case "ok && newGen != oldGen +1"? Should that be handled?

pkg/reconciler/stats_reporter.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const (
+	knativeServingChangeCountName = "knativeserving_change_count"
+)
+
+var (
+	knativeServingChangeCountStat = stats.Int64(
+		knativeServingChangeCountName,
+		"Number of KnativeServing changes including creation, edit and deletion",
```
anniefu: [nit] "Number of changes to the KnativeServing Custom Resource where a change can be create, update, or delete."

anniefu:
Could you test with stackdriver by setting these in config-observability?
```
    metrics.backend-destination: stackdriver
    metrics.request-metrics-backend-destination: stackdriver
    metrics.allow-stackdriver-custom-metrics: "true"
```
Otherwise, 
/lgtm

jihuin:
@anniefu Tested with stackdriver, and worked as expected.

pkg/reconciler/knativeserving/knativeserving_controller.go
jihuin: Attaching the generation may be not that useful for the metrics only counting the number of changes.

PR #237, opened 2019-11-29
Feature: reuse e2e tests on productized code

cardil:
## Proposed Changes

* Rely on PR #223
* Expose e2e test as a compliance suite (dbfbf2c)
* Add a test context to fine tune compliance suite testing (5d95dd6)

#### Expose e2e test as a compliance suite (dbfbf2c)

As a knative vendor we would like to execute tests from upstream knative in out productized code's test plans.

To do that, I propose this minor change that would result in type safe way of reusing upstream test code, by vendoring it in our code. It involves moving e2e test to regular `.go` file, instead of `_test.go` file. In that way we can call that code from our test suite.

#### Add a test context to fine tune compliance suite testing (5d95dd6)

Sometimes upstream tests could not pass yet. That's natural as vendors will always be behind upstream code. In that case we need a mechanism that will let us skip parts of the tests.

That was implemented in `test.Context` object.

### Example usage

There is already created a PR openshift-knative/serverless-operator#32 that uses this concept. Below I highlighted a key test file from that PR (note that we are using [0.9 version](https://github.com/cardil/serving-operator/tree/feature/v0.9-reuse-tests-on-productized) of this change):

```go
package e2e

import (
	"testing"

	"github.com/openshift-knative/serverless-operator/test"
	upstreame2e "knative.dev/serving-operator/test/e2e"
	upstreamtest "knative.dev/serving-operator/test"
)

func TestUpstreamKnativeServingOperator(t *testing.T) {
	upstreamtest.ServingOperatorNamespace = "knative-serving"
	suite := upstreame2e.ComplianceSuite()
	ctx := test.SetupClusterAdmin(t)

	defer test.CleanupAll(ctx)
	test.CleanupOnInterrupt(t, func() { test.CleanupAll(ctx) })

	createSubscriptionAndWaitForCSVtoSucceed(t, ctx) // <1>

	skipConfigureSubTest := upstreamtest.Skip(
		"TestKnativeServingDeployment/configure",
		"Skip due to SRVKS-241") // <2>

	upstreamtest.
		NewContext(t).
		WithOverride(skipConfigureSubTest).
		RunSuite(suite) // <3>

	undeployServerlessOperatorAndCheckDependentOperatorsRemoved(t, ctx) // <4>
}
```

* `<1>` setup of our Serverless operator
* `<2>` registering a skip for `TestKnativeServingDeployment/configure` that we are not supporting at this point
* `<3>` actual compliance suite execution
* `<4>` teardown of our Serverless operator

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Temporary Build Files
 build/_output
 build/_test
+# Goland
+.idea
+
 # Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Emacs ###
 # -*- mode: gitignore; -*-
--- a/cmd/manager/kodata/knative-serving/dummy.go
+++ b/cmd/manager/kodata/knative-serving/dummy.go
@@ -1,11 +1,12 @@
-// +build e2e
-
 /*
 Copyright 2019 The Knative Authors
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -13,13 +14,5 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package e2e
-
-import (
-	"testing"
-)
-
-// TestKnativeServingDeployment verifies the KnativeServing creation, deployment recreation, and KnativeServing deletion.
-func TestKnativeServingDeployment(t *testing.T) {
-	testKnativeServingDeployment(t)
-}
+// Package knative_serving is a dummy buildable source file to satisfy `dep`
+package knative_serving
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -20,13 +20,13 @@ set -o pipefail
 
 source $(dirname $0)/../vendor/knative.dev/test-infra/scripts/library.sh
 
-cd ${REPO_ROOT_DIR}
+cd "${REPO_ROOT_DIR}"
 
 # Ensure we have everything we need under vendor/
 dep ensure
 
-rm -rf $(find vendor/ -name 'OWNERS')
-rm -rf $(find vendor/ -name '*_test.go')
+find vendor/ -name 'OWNERS' -delete
+find vendor/ -name '*_test.go' -delete
 
 update_licenses third_party/VENDOR-LICENSE "./cmd/*"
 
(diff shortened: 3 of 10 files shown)


test/e2e/knativeservingdeployment_test.go
```diff
@@ -16,241 +16,12 @@ limitations under the License.
...
-		return err
-	}
-	if len(servings.Items) > 0 {
-		return errors.New("Unable to verify cluster-scoped resources are deleted if any KnativeServing exists")
-	}
-	return nil
-}
+	ctx := test.NewContext(t)
+	testKnativeServingDeployment(ctx)
+}
```
mattmoor: Add trailing newline:
```suggestion
}

```

markusthoemmes:
I do think this concept spans wide enough to be proposed and discussed in the productivity WG. It only really makes sense if something like this is adopted or at least acknowledged across the entirety of Knative.

markusthoemmes:
What's also not entirely clear to me yet: Why is this better than just using `go test`?

markusthoemmes:
To further clarify (based on a Slack discussion), referencing the bullets in your example.

1. This is in the setup scripts of this repository. I think if we want to adopt this, this repository should also **deploy** the operator in the same vain rather than using scripts. Since we want to deploy our own operator as a vendor, that should then be just a modification of the same steps. I don't know if I agree on having that be part of the tests themselves though.
2. Is arguably nice to have but should imo be made part of `go test`.
3. Is just `go test` currently.
4. Same as in 1., this is currently part of the scripts.

cardil:
@markusthoemmes proposing another way of reusing tests. That's to create a bash script that create test context and execute tests.

My approach is different. Writing everything as a Go test will be better of ways:

* It will be easier to understand and write properly as bash isn't well understood and written well
* test reports will be the same as in other tests
* It will be easier to reuse in different tests
* It's easier to select what tests should be run, with build tags
* In future we might rewrite cluster setup to GO. Then, we will be able to run and DEBUG single test directly from IDE

The last point is biggest for me. That's what I would like to see. To be able to run everything just from IDE. Just as you run Spring Test or Arquillian!

If you think of e2e testing, that might be written in every language that you want. Any! That's that it simulates user operations on live system.

But, writing those in the same language as production code and unit tests code is beneficial. You can use fewer languages to do the job. You are more proficient with language that you're writing the production code in and can use the same tools to do linting, static analysis.

cardil:
>     2. Is arguably nice to have but should imo be made part of `go test`.

You can't exclude tests with `go test`. It's only additive.

adrcunha:
Related: https://github.com/knative/serving/issues/4263

chaodaiG:
Is this still being worked on?

coryrc:
Possibly related: knative/serving#4263

cardil:
This is PoC of solution to be discussed.

aliok:
@cardil how should we proceed with this one? anything I can do to help?

cardil:
It should be discussed on TOC meeting. It was already pitched there.

houshengbo:

PR #254, opened 2020-01-16
Update the year of the copyright to 2020 in the template

houshengbo:
## Proposed Changes

* As we move into the year of mouse, update the year to 2020 in the template.

**Release Note**

```release-note

```

### DIFF
--- a/hack/boilerplate/boilerplate.go.txt
+++ b/hack/boilerplate/boilerplate.go.txt
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Knative Authors
+Copyright 2020 The Knative Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


markusthoemmes:
We don't usually do this in other repos. Is this required?

houshengbo:
/hold
Let me do a confirmation about the change.

houshengbo:
/hold cancel
@markusthoemmes You are correct. The source code should stay with the original, so only change the year in the template.

Here is some discussions: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

In general, there are two type voices:
1. Stay with the year when the work is first published.
2. We CAN update a copyright year only if contributions are made to the work during that year.

k4leung4:
/lgtm

it does seem we do this for serving and eventing
https://github.com/knative/serving/pull/6464
https://github.com/knative/eventing/pull/2359

PR #255, opened 2020-01-16
Add the template for Knative projects

houshengbo:


aliok:
@houshengbo I think while at it, we should also have the PR template, like this: https://github.com/knative/eventing/blob/master/.github/pull-request-template.md

PR #257, opened 2020-01-20
Remove all the deployments in the older version

### LINKED ISSUE #256: Remove the resources of the older version, if a new version of knative is installed.
Currently, when we install the operator of an new version, in a cluster, where there is an older version operator, the resources of knative serving will be overwritten or brand-new installed, depending on whether the resource exists or not.

However, if the new version of knative serving remove resources compared to the old version, the operator should remove those resources.

Since all the resources are labeled with "serving.knative.dev/release="\<Version\>", we can use the version number to find all the resources at the older version.

houshengbo:
Fixes #256

## Proposed Changes

* We can know if there is an older version of operator CR, by checking Status.Version. This PR add the support to delete the old resources with the label "serving.knative.dev/release=\<oldVersion\>", If it is not empty and not equal to the current version.

**Release Note**

```release-note

```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -45,6 +45,7 @@ const (
 	creationChange = "creation"
 	editChange     = "edit"
 	deletionChange = "deletion"
+	selectorKey    = "serving.knative.dev/release"
 )
 
 var (
@@ -59,6 +60,7 @@ type Reconciler struct {
 	knativeServingLister listers.KnativeServingLister
 	config               mf.Manifest
 	servings             map[string]int64
+	oldVersion           string
 }
 
 // Check that our Reconciler implements controller.Reconciler
@@ -105,6 +107,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
 		}
 	}
 	r.servings[key] = newGen
+
+	// Save the version of the original CR
+	r.oldVersion = original.Status.Version
 
 	// Don't modify the informers copy.
 	knativeServing := original.DeepCopy()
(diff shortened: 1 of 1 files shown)


markusthoemmes:
What is the expected behavior? Is this supposed to remove „outdated“ deployments i.e. those that are no longer used in the new version?

houshengbo:
@markusthoemmes Correct. I am also seeking to remove some other older resources, which are not used any more in the new version, like service, configmaps, gateway, etc, but I did not find the solution to remove all of them will a single call.

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -295,5 +300,26 @@ func (r *Reconciler) deleteObsoleteResources(manifest *mf.Manifest, instance *se
...
+	// the label serving.knative.dev/release="<oldVersion>"
+	if r.oldVersion != "" && r.oldVersion != version.Version {
+		selector := fmt.Sprintf("%s=%s", selectorKey, r.oldVersion)
+		listOptions := metav1.ListOptions{ LabelSelector: selector }
+
+		// Remove all the deployments at the old version
+		deploymentList, err := r.KubeClientSet.AppsV1().Deployments(instance.GetNamespace()).List(listOptions)
+		if err == nil {
+			for _, dep := range deploymentList.Items {
+				if e := r.KubeClientSet.AppsV1().Deployments(instance.GetNamespace()).Delete(dep.Name,
```
jcrossley3: This makes me nervous. Deleting a deployment instead of updating it introduces the potential for losing requests, I think. Besides, I prefer seeing in this function exactly which resources are obsolete and need to be deleted.

houshengbo: @jcrossley3 To find out all the resources to be removed, I can think of using the manifestival package to read the manifest of the older version, and compare to the manifest of the new version, to see if they still exist in the new version. Do you have any other suggestions?

houshengbo:
We can use an alternative way to delete the obsolete resources with the existing `deleteObsoleteResources` function: when there is a new release of knative serving, check if there are resources deleted in the new release. If so, we add delete calls to those resources in the function `deleteObsoleteResources`.

Based on the nature of the low frequency that a resource is deleted, we can do easy diff to find out the missing resource and maintain `deleteObsoleteResources`. It is ideal to delete all resources that don‘t match the current version, but that brings in too much overhead.

Issue #280, opened 2020-02-02
Extend image override logic to Daemonsets


jcrossley3:
**Describe the bug**
The forthcoming knative-serving release will deploy the activator as a daemonset, but our image override logic only replaces images in `Deployments` currently.

**Expected behavior**
Images for Daemonsets should be subject to the same override logic as for Deployments

**To Reproduce**
Try overriding the activator images and see that its Daemonset image changes.

**Knative release version**
Post 0.12

**Additional context**
This is the change upstream: https://github.com/knative/serving/pull/6624

This issue might lend some justification to #277 as it would be handy to be able to test a released operator with an unreleased manifest.

PR #281, opened 2020-02-03
Add image override logic for DaemonSet

savitaashture:
Fixes https://github.com/knative/serving-operator/issues/280

## Proposed Changes

* Extended image override logic for DaemonSet when there is image update by [override](https://github.com/savitaashture/serving-operator/blob/master/pkg/apis/serving/v1alpha1/knativeserving_types.go#L47) 

**Release Note**

This PR will be validated when release manifest updated to changes from knative/serving#6624
This PR has dependency [knative.dev/pkg](https://github.com/knative/pkg/pull/1043)

### DIFF
--- a/pkg/reconciler/knativeserving/common/extensions.go
+++ b/pkg/reconciler/knativeserving/common/extensions.go
@@ -32,7 +32,6 @@ func (platforms Platforms) Transformers(kubeClientSet kubernetes.Interface, inst
 		mf.InjectOwner(instance),
 		mf.InjectNamespace(instance.GetNamespace()),
 		ConfigMapTransform(instance, log),
-		ResourceTransform(instance, log),
 		ImageTransform(instance, log),
 		GatewayTransform(instance, log),
 		CustomCertsTransform(instance, log),
--- a/pkg/reconciler/knativeserving/common/extensions.go
+++ b/pkg/reconciler/knativeserving/common/extensions.go
@@ -32,7 +32,7 @@ func (platforms Platforms) Transformers(kubeClientSet kubernetes.Interface, inst
 		mf.InjectOwner(instance),
 		mf.InjectNamespace(instance.GetNamespace()),
 		ConfigMapTransform(instance, log),
-		DeploymentTransform(instance, log),
+		ResourceTransform(instance, log),
 		ImageTransform(instance, log),
 		GatewayTransform(instance, log),
 		CustomCertsTransform(instance, log),
--- a/pkg/reconciler/knativeserving/common/images.go
+++ b/pkg/reconciler/knativeserving/common/images.go
@@ -41,15 +41,14 @@ var (
 // ResourceTransform updates deployment and daemonSet with a new registry and tag
 func ResourceTransform(instance *servingv1alpha1.KnativeServing, log *zap.SugaredLogger) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
-		// Update the deployment with the new registry and tag
-		if u.GetKind() == "Deployment" {
+		switch u.GetKind() {
+		case "Deployment":
 			return updateDeployment(instance, u, log)
-		}
-		// Update the daemonSet with the new registry and tag
-		if u.GetKind() == "DaemonSet" {
+		case "DaemonSet":
 			return updateDaemonSet(instance, u, log)
+		default:
+			return nil
 		}
-		return nil
 	}
 }
 
--- a/pkg/reconciler/knativeserving/common/images.go
+++ b/pkg/reconciler/knativeserving/common/images.go
@@ -51,24 +51,20 @@ func ImageTransform(instance *servingv1alpha1.KnativeServing, log *zap.SugaredLo
 			if u.GetAPIVersion() == "caching.internal.knative.dev/v1alpha1" {
 				return updateCachingImage(instance, u)
 			}
-			return nil
-		default:
-			return nil
 		}
+		return nil
 	}
 }
 
 func updateDeployment(instance *servingv1alpha1.KnativeServing, u *unstructured.Unstructured, log *zap.SugaredLogger) error {
 	var deployment = &appsv1.Deployment{}
-	err := scheme.Scheme.Convert(u, deployment, nil)
-	if err != nil {
+	if err := scheme.Scheme.Convert(u, deployment, nil); err != nil {
 		log.Error(err, "Error converting Unstructured to Deployment", "unstructured", u, "deployment", deployment)
 		return err
 	}
 
 	updateRegistry(&deployment.Spec.Template.Spec, instance, log, deployment.GetName())
-	err = scheme.Scheme.Convert(deployment, u, nil)
-	if err != nil {
+	if err := scheme.Scheme.Convert(deployment, u, nil); err != nil {
 		return err
 	}
 	// The zero-value timestamp defaulted by the conversion causes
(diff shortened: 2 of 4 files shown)


pkg/reconciler/knativeserving/common/images.go
```diff
@@ -38,12 +38,17 @@ var (
...
 
-func DeploymentTransform(instance *servingv1alpha1.KnativeServing, log *zap.SugaredLogger) mf.Transformer {
+// ResourceTransform updates deployment and daemonSet with a new registry and tag
+func ResourceTransform(instance *servingv1alpha1.KnativeServing, log *zap.SugaredLogger) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		// Update the deployment with the new registry and tag
 		if u.GetKind() == "Deployment" {
 			return updateDeployment(instance, u, log)
 		}
+		// Update the daemonSet with the new registry and tag
```
markusthoemmes: nit:

```golang
switch u.GetKind()
case "Deployment":
  return updateDeployment(instance, u, log)
case "DaemonSet":
  return updateDaemonSet(instance, u, log)
default:
  return nil
}
```

pkg/reconciler/knativeserving/common/extensions.go
```diff
@@ -32,7 +32,7 @@ func (platforms Platforms) Transformers(kubeClientSet kubernetes.Interface, inst
 		mf.InjectOwner(instance),
 		mf.InjectNamespace(instance.GetNamespace()),
 		ConfigMapTransform(instance, log),
-		DeploymentTransform(instance, log),
+		ResourceTransform(instance, log),
```
jcrossley3: I think name this is a little vague. DaemonSets and StatefulSets are really just specialized Deployments. And pretty much everything is a Resource. I'd say just keep the name `DeploymentTransform` or maybe `PodSpecableTransform`? Or save that for another PR?

savitaashture:
> PodSpecable

Yep true i will make use of `PodSpecable` in the PR and update

savitaashture:
> I almost feel like it's worth biting the bullet and using `PodSpecable`. Seems like that would reduce a lot of the code in the PR and simultaneously pull in support for `StatefulSets` as well. WDYT?

Yep true i will make use of `PodSpecable` in the PR and update

pkg/reconciler/knativeserving/common/extensions.go
savitaashture: Sure i will keep `PodSpecableTransform`  :+1:

pkg/reconciler/knativeserving/common/images.go
savitaashture: Thank you @markusthoemmes Updated PR

savitaashture:
CI fail until [PR](https://github.com/knative/pkg/pull/1043) merge

pkg/reconciler/knativeserving/common/images.go
```diff
@@ -20,60 +20,62 @@ import (
...
 
-	updateDeploymentImage(deployment, &registry, log)
-	deployment.Spec.Template.Spec.ImagePullSecrets = addImagePullSecrets(
-		deployment.Spec.Template.Spec.ImagePullSecrets, &registry, log)
-	err = scheme.Scheme.Convert(deployment, u, nil)
+	updateDeploymentImage(specData.Spec.Template.Spec.Containers, specData.GetName(), &registry, log)
+	specData.Spec.Template.Spec.ImagePullSecrets = addImagePullSecrets(
+		specData.Spec.Template.Spec.ImagePullSecrets, &registry, log)
+
+	unstructuredData, err := duck.ToUnstructured(specData.DeepCopyObject().(duck.OneOfOurs))
```
markusthoemmes: Oh I see, you need the `selector` in PodSpecable to successfully serialize back here? 🤔 

Can you instead use the existing unstructured and only override things? I.e. take the transformed podSpec and set it into the unstructured element at `Spec.Template` or something like that?

pkg/reconciler/knativeserving/common/images.go
```diff
@@ -20,60 +20,62 @@ import (
...
-}
-
+// ImageTransform updates image with a new registry and tag
 func ImageTransform(instance *servingv1alpha1.KnativeServing, log *zap.SugaredLogger) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
-		// Update the image with the new registry and tag
-		if u.GetAPIVersion() == "caching.internal.knative.dev/v1alpha1" && u.GetKind() == "Image" {
-			return updateCachingImage(instance, u)
+		switch u.GetKind() {
+		case "Deployment", "DaemonSet":
```
jcrossley3: Wanna knock out "StatefulSet" while we're at it? :smile:

pkg/reconciler/knativeserving/common/images.go
savitaashture: Do you mean this https://github.com/knative/serving-operator/pull/281/files#diff-411a842e3f404ed5adc1b082ea595d72R78 directly instead of duck.ToUnstructured ?

markusthoemmes: No I don't think that'd work either. Lemme try if what I though actually works before I bother you more with it 😂

savitaashture: Yep that also won't work

markusthoemmes: In pseudo-code, we need something like this:

```
podSpecable := duck.FromUnstructured(u)

updateImages(podSpecable)
updatePullSecret(podSpecable)

unstructured.SetNestedField(u.Object, podSpecable.Spec.Template, "spec", "template")
```

Or alternatively

```
podSpecable := duck.FromUnstructured(u)
afterChanges := podSpecable.DeepCopy()

updateImages(afterChanges)
updatePullSecret(afterChanges)

patch := generatePatch(podSpecable, afterChanges)
applyPatch(u, patch)
```

That way we keep the fields of the unstructured intact that PodSpecable doesn't capture (see selector etc.) while also having the profit of using just one type. I haven't gotten a working version of the above together yet though.

pkg/reconciler/knativeserving/common/images.go
```diff
@@ -38,23 +38,23 @@ var (
...
+			return updateDaemonSet(instance, u, log)
+		case "Image":
+			if u.GetAPIVersion() == "caching.internal.knative.dev/v1alpha1" {
+				return updateCachingImage(instance, u)
+			}
+			return nil
+		default:
+			return nil
 		}
-		return nil
```
jcrossley3: nit: i would lose the default case and replace the two returns in the switch with the old one at the end of the function.

pkg/reconciler/knativeserving/common/images.go
```diff
@@ -84,23 +79,51 @@ func updateDeployment(instance *servingv1alpha1.KnativeServing, u *unstructured.
...
-// updateDeploymentImage updates the image of the deployment with a new registry and tag
-func updateDeploymentImage(deployment *appsv1.Deployment, registry *servingv1alpha1.Registry, log *zap.SugaredLogger) {
-	containers := deployment.Spec.Template.Spec.Containers
+func updateDaemonSet(instance *servingv1alpha1.KnativeServing, u *unstructured.Unstructured, log *zap.SugaredLogger) error {
+	var daemonSet = &appsv1.DaemonSet{}
+	err := scheme.Scheme.Convert(u, daemonSet, nil)
+	if err != nil {
+		log.Error(err, "Error converting Unstructured to daemonSet", "unstructured", u, "daemonSet", daemonSet)
+		return err
+	}
```
jcrossley3: nit: do the assignment in the if clause for consistency

pkg/reconciler/knativeserving/common/images.go
```diff
@@ -84,23 +79,51 @@ func updateDeployment(instance *servingv1alpha1.KnativeServing, u *unstructured.
...
+	updateRegistry(&daemonSet.Spec.Template.Spec, instance, log, daemonSet.GetName())
+	err = scheme.Scheme.Convert(daemonSet, u, nil)
+	if err != nil {
+		return err
+	}
```
jcrossley3: nit: same as ^^

savitaashture:
@jcrossley3 thank you
Addressed all the comments

jcrossley3:
Thanks @savitaashture :smile: 
/lgtm
/approve

Issue #282, opened 2020-02-03
Principle of least privilege should be applied in operator (Cluster)Roles


Cynocracy:
**Describe the bug**
Today the operator is by default granted all permissions on all resources clusterwide by a [blanket ClusterRole](https://github.com/knative/serving-operator/blob/master/config/role.yaml#L60)

**Expected behavior**
Only specifically those permissions which are necessary in order to create a functional Knative serving setup should be added, using [escalate and/or bind](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping) to create roles or rolebindings that the operator itself does not need explicitly, but which are needed transitively by knative serving.

**To Reproduce**
1. Install Knative Serving Operator.
1. kubectl describe clusterrole knative-serving-operator
```
Name:         knative-serving-operator
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"rbac.authorization.k8s.io/v1","kind":"ClusterRole","metadata":{"annotations":{},"name":"knative-serving-operator"},"rules":...
PolicyRule:
  Resources  Non-Resource URLs  Resource Names  Verbs
  ---------  -----------------  --------------  -----
  *.*        []                 []              [*]
```

jcrossley3:
The primary reason the operator is given `cluster-admin` privs is because the knative-serving release manifest employs aggregated clusterroles. At the time, this feature required `cluster-admin` privs. I'm not sure if that's still the case or not. /cc @pmorie

Cynocracy:
Thanks for the context! That makes sense.

It looks like we may be able to have the Operator config set up the aggregated clusterrole, and grant the Operator service account `bind` on that role.

pmorie:
hi @Cynocracy, unfortunately you need */* permissions in order to use cluster role aggregation:

https://github.com/kubernetes/kubernetes/blob/3b618af0d435628feedf06f97bd1c69340d07d95/pkg/registry/rbac/clusterrole/policybased/storage.go#L79-L83

Cynocracy:
Thanks for the link, I'm still very new to the space so it's quite helpful for me :)

For now, I'll start a PR adding a comment to the `*/*` permission. It'd be nice if we can get some kind of fix upstream that removes the req, but at least with the current setup, I found the very explicit list followed by `*/*` to be quite confusing.

Cynocracy:
Circling back to this, a couple of different things:

[...]

PR #284, opened 2020-02-04
Add a comment about why cluster-admin is necessary.

Cynocracy:
Unfortunately it appears that in order to use an aggregated
clusterrole we need the operator to be a cluster-admin.
This should help prevent confusion re: why we have both an
explicit list and a blanket `*/*` granted to the operator
service account.

See #282 

## Proposed Changes

* Add a (hopefully) helpful comment.

**Release Note**

NONE


Cynocracy:
/assign @greghaynes

(Is it normal to have separate LGTMer and approver? :)

Cynocracy:
Thank you!

PR #291, opened 2020-02-11
Restrict RBAC for Operator

### LINKED ISSUE #282: Principle of least privilege should be applied in operator (Cluster)Roles

Cynocracy:
Fixes #282

#374  is a base of this, you can solely review the YAML here
(and not the reconciler change) though you're welcome to review
that as well.

## Proposed Changes

By using the `escalate` verb on clusterroles, we allow the Operator
to have the ability to create any clusterrole. This prevents us from
needing to grant it `cluster-admin`.

Additionally, by having the Operator create an aggregated clusterrole
for itself (aggregating all clusteroles created by Knative Serving),
we pick up any permissions that the upstream dictate as necessary
by any component.

Any roles that are not created by upstream, but which are bound as
part of install are added with the `bind` verb to allow the Operator
to bind them without necessarily having the permissions they grant.

Some other permissions are needed to support controller infra
and are explicitly included, as well as some specific to the Operator
itself. An attempt at grouping these by sections is included, though
may be slightly out of sync (in particular: Controller infra / Specific
to this Operator may overlap).

Overall, this should keep the Operator roughly operating only on
Knative Serving resources (or things which can be operated
on by upstream).

**Release Note**

```release-note
Change RBAC to minimize permissions of Operator service account.
```

### DIFF
--- a/config/role.yaml
+++ b/config/role.yaml
@@ -11,23 +11,99 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
-  creationTimestamp: null
+  name: knative-serving-operator-aggregated
+aggregationRule:
+  clusterRoleSelectors:
+# This (along with escalate below) allows the Operator to pick up any
+# roles that are provided to the admin of the cluster by knative serving
+# automatically.
+  - matchExpressions:
+      - {key: serving.knative.dev/release, operator: Exists}
+rules: [] # Rules are automatically filled in by the controller manager.
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
   name: knative-serving-operator
 rules:
 - apiGroups:
+  - operator.knative.dev
+  resources:
+  - '*'
+  verbs:
+  - '*'
+# Bootstrapping permissions.
+# Roles that are explicitly bound buch which are specified by this Operator
+# MUST be specified here with 'get' and 'bind'.
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterroles
+  resourceNames:
+  - system:auth-delegator
+  verbs:
+  - bind
+  - get
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  resourceNames:
+  - extension-apiserver-authentication-reader
+  verbs:
+  - bind
+  - get
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterroles
+  - roles
+  verbs:
+  - create
+  - delete
+# Escalate is necessary in order to create a role using cluster role aggregation,
+# and to allow the Operator to bootstrap itself into the necessary set of
+# permissions, even as those continue to evolve upstream.
+  - escalate
+  - get
+  - list
+  - update
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterrolebindings
+  - rolebindings
+  verbs:
+  - create
+  - delete
+  - list
+  - get
+  - update
+# Permissions required for Knative controller
+# infra.
+- apiGroups:
+  - apiregistration.k8s.io
+  resources:
+  - apiservices
+  verbs:
+  - update
+- apiGroups:
   - ""
   resources:
-  - pods
   - services
-  - endpoints
-  - persistentvolumeclaims
-  - events
-  - configmaps
-  - secrets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
+- apiGroups:
+  - caching.internal.knative.dev
+  resources:
+  - images
   verbs:
   - '*'
 - apiGroups:
@@ -36,6 +112,26 @@ rules:
   - namespaces
   verbs:
   - get
+  - update
+  - watch
+- apiGroups:
+  - ''
+  resources:
+  - events
+  verbs:
+  - create
+  - update
+  - patch
+- apiGroups:
+  - ''
+  resources:
+  - configmaps
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
 - apiGroups:
   - apps
   resources:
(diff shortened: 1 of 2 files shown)


config/role.yaml
```diff
@@ -22,43 +103,86 @@ rules:
...
+- apiGroups:
+  - autoscaling
+  resources:
+  - horizontalpodautoscalers
+  verbs:
   - create
+  - delete
+  - get
+  - list
+# Permissions required specifically for this Operator.
```
mattmoor: Remove trailing whitespace:
```suggestion
# Permissions required specifically for this Operator.
```

Cynocracy:
/retest

(seems to be running OK locally)

Cynocracy:
Oh, I see the issue.

Cynocracy:
/retest

:(

Cynocracy:
Found out how to read the logs:

{"ts":"2020-02-12T01:09:32.279Z","knative.dev/controller":"knativeserving-controller","msg":"Reconcile succeeded. Time taken: 11.99198359s.","commit":"0b4d6f8","level":"info","logger":"serving_operator.knativeserving-controller","knative.dev/key":"operator-tests/knative-serving","caller":"controller…

Hmmmm

/retest

so that I can poke around more :)

Cynocracy:
raceid":"fd80c601-0e62-4773-bdd6-1ecb60dbbf78","knative.dev/key":"operator-tests/knative-serving"}
{"level":"info","ts":"2020-02-13T01:16:24.454Z","logger":"serving_operator.knativeserving-controller.event-broadcaster","caller":"record/event.go:274","msg":"Event(v1.ObjectReference{Kind:\"KnativeServing\", Namespace:\"operator-tests\", Name:\"knative-serving\", UID:\"5e9ab652-b2da-431a-86cb-211efa5cbd70\", APIVersion:\"operator.knative.dev/v1alpha1\", ResourceVersion:\"4761\", FieldPath:\"\"}): type: 'Warning' reason: 'InternalError' apiservices.apiregistration.k8s.io \"v1beta1.custom.metrics.k8s.io\" is forbidden: User \"system:serviceaccount:default:knative-serving-operator\" cannot update resource \"apiservices\" in API group \"apiregistration.k8s.io\" at the cluster scope","commit":"b29f72a","knative.dev/controller":"knativeserving-controller"}

OK, so that was required.

Cynocracy:
This is ready for a preliminary review.

Let me know if more detail is necessary.

config/role.yaml
```diff
@@ -4,16 +4,103 @@ metadata:
...
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterrolebindings
+  - rolebindings
+  verbs:
+  - create
+  - delete
+  - get
+  - update
+# Permissions required for Knative controller
```
trshafer: Should the verbs here be escalate?

config/role.yaml
```diff
@@ -22,43 +109,86 @@ rules:
...
+- apiGroups:
+  - ''
+  resources:
+  - nodes
   resourceNames:
-  - knative-serving-operator
+  - minikube
+  verbs:
+  - get
+# Old resources that need cleaning up that are not in the knative-serving
```
trshafer: When can these be deleted?

config/role.yaml
```diff
@@ -4,16 +4,103 @@ metadata:
...
+# automatically.
+  - matchExpressions:
+      - {key: serving.knative.dev/release, operator: Exists}
+rules: [] # Rules are automatically filled in by the controller manager.
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: knative-serving-operator
+rules:
```
trshafer: Are these rules manually curated?

config/role.yaml
Cynocracy: Whenever the Operator code to clean up the resources can be, it's not clear to me what our strategy is in that regard.

I imagine it will become clear once we have clear upgrade/downgrade and versioning story between the operator what it installs :)

config/role.yaml
Cynocracy: Yes.

In general, only the `bind` section should need modification (if we add new non-Knative managed roles to the rolebindings used in Serving)

config/role.yaml
Cynocracy: In RoleBindings?

The equivalent for `escalate` for bindings is `bind` (which allows you to bind a role that grants permissions you do not have). There are a few explicit `bind` grants above, maybe it would be clearer if I move those closer to these other rolebinding permissions?

Cynocracy:
I think splitting it into multiple files makes sense, especially since some of the sections will be drop-in equivalents between the serving and eventing operator.

I'm working on another WIP for that repo, let me take a stab at splitting once I have that working (it will be a good proof of concept that the general idea of 'escalate' + 'bind' maps well).

Cynocracy:
{"level":"info","ts":"2020-03-06T23:30:27.273Z","logger":"serving-operator.knativeserving-controller.event-broadcaster","caller":"record/event.go:274","msg":"Event(v1.ObjectReference{Kind:\"KnativeServing\", Namespace:\"knative-serving\", Name:\"knative-serving\", UID:\"cbacee1e-5ed9-4bac-a8ba-d65b0ba9225a\", APIVersion:\"operator.knative.dev/v1alpha1\", ResourceVersion:\"2288\", FieldPath:\"\"}): type: 'Warning' reason: 'InternalError' serviceaccounts \"controller\" is forbidden: User \"system:serviceaccount:default:knative-serving-operator\" cannot get resource \"serviceaccounts\" in API group \"\" in the namespace \"knative-serving\"","commit":"1419ff1","knative.dev/controller":"knativeserving-controller"}

Cynocracy:
Wait.. it should have * * * in knative-serving.

trshafer:
> Wait.. it should have * * * in knative-serving.

Are there more error messages or does it need GET serviceaccounts in knative-serving?

Cynocracy:
I believe the issue has been fixed (the issue was that the role and the rolebinding need to be in the relevant namespace, and the subjects needed to point to the default namespace).

This opens up a question I need to resolve though: do we need to support multiple KS installations in the same cluster, or can we assume the default namespace is used?

Cynocracy:
OK, the error this time: deleting fails, because we delete the role that gives us the ability to delete 'admissionregistration.k8s.io' resources before we delete those resources.

Going to take a stab at saving the deletion of the roles/rolebindings until *last* as we need to 'unbootstrap' ourselves from having these permissions.

Cynocracy:
WIP as eventing change still is, and I'd like these to go in concurrently if at all.

Cynocracy:
/assign @jcrossley3

Assuming my last minute diff-tweaking didn't break this, this should be ready to review :)

The only remaining diff between the role here and in eventing are the replacement of some labels,
a few explicit 'bind's (which now need to be added to roles the Operator binds to but
does not create), and the resource deletion section at the end.

Cynocracy:
One more comment re: the last deltas.

In the future, if we move to a job based cleanup, deletion roles can be defined one-off.
If the Operator repos and/or controllers are to merge, this dramatically simplifies the collapse of 
the roles needed.

config/role.yaml
Cynocracy: See comment on manual curation thread.

config/role.yaml
Cynocracy: Copied from below:

The only remaining diff between the role here and in eventing are the replacement of some labels,
a few explicit 'bind's (which now need to be added to roles the Operator binds to but
does not create), and the resource deletion section at the end.

In the future, if we move to a job based cleanup, deletion roles can be defined one-off.
If the Operator repos and/or controllers are to merge, this dramatically simplifies the collapse of
the roles needed.

config/role.yaml
Cynocracy: See manually curated comment chain.

Cynocracy:
Commits now squashed (squished?) into a single one :)

Thanks for the tip jcrossley3

jcrossley3:
Runs fine on OpenShift, so...
/lgtm

Cynocracy:
To double check the diff, I did

1. ko apply config 
2. kubectl api-resources -oname | xargs -I% kubectl get -A % -oname | sort | uniq > /tmp/before
3. Create the KS resource
4. Wait for it to finish
5. Delete the KS resource
6. kubectl api-resources -oname | xargs -I% kubectl get -A % -oname | sort | uniq > /tmp/after
7 diff /tmp/before /tmp/after

Which spat out

https://pastebin.com/tugjzdeU

The only oddness is the activator pod, which seems to be hanging around.

Cynocracy:
https://pastebin.com/bjuPParB

Before, no activator leak, curious.

Cynocracy:
A second attempt at a diff produced https://pastebin.com/N0w4R9QZ

So it's entirely possible I generated the last one while the pod was going down (I only diff the names, so who knows :)

Either way, this is ready for an approval, IMO.

PR #293, opened 2020-02-12
Support reading KubeConfig from KUBECONFIG env var

### LINKED ISSUE #292: Support reading KubeConfig from KUBECONFIG env var
Same issue as eventing one, https://github.com/knative/eventing-operator/issues/99

aliok:
Fixes #292 

## Proposed Changes

* Backwards compatible change
* Only use `KUBECONFIG` if the command line argument is empty

**Release Note**

```release-note
NONE
```

Same PR for eventing operator: https://github.com/knative/eventing-operator/pull/100

### DIFF
--- a/pkg/reconciler/knativeserving/controller.go
+++ b/pkg/reconciler/knativeserving/controller.go
@@ -16,6 +16,7 @@ package knativeserving
 import (
 	"context"
 	"flag"
+	"knative.dev/pkg/injection/sharedmain"
 	"os"
 	"path/filepath"
 
@@ -23,7 +24,6 @@ import (
 	mf "github.com/manifestival/manifestival"
 	"go.uber.org/zap"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/clientcmd"
 	deploymentinformer "knative.dev/pkg/client/injection/kube/informers/apps/v1/deployment"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
@@ -60,7 +60,7 @@ func NewController(
 
 	koDataDir := os.Getenv("KO_DATA_PATH")
 
-	cfg, err := clientcmd.BuildConfigFromFlags(*MasterURL, *Kubeconfig)
+	cfg, err := sharedmain.GetConfig(*MasterURL, *Kubeconfig)
 	if err != nil {
 		c.Logger.Error(err, "Error building kubeconfig")
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -19,7 +19,6 @@ import (
 	"flag"
 	"log"
 
-	"k8s.io/client-go/tools/clientcmd"
 	"knative.dev/pkg/injection/sharedmain"
 	"knative.dev/pkg/signals"
 	"knative.dev/serving-operator/pkg/reconciler/knativeserving"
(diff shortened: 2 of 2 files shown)


pkg/reconciler/knativeserving/controller.go
```diff
@@ -40,7 +40,7 @@ const (
 var (
 	recursive  = flag.Bool("recursive", false, "If filename is a directory, process all manifests recursively")
 	MasterURL  = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
-	Kubeconfig = flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	Kubeconfig = flag.String("kubeconfig", os.Getenv("KUBECONFIG"), "Path to a kubeconfig. Only required if out-of-cluster.")
```
markusthoemmes: Can we actually remove these altogether? They should come via sharedmain anyway, I think? (modulo recursive)

houshengbo: How can we access these arguments from sharedmain?
```
config, err := mf.NewManifest(filepath.Join(koDataDir, "knative-serving/"), *recursive, cfg)
```
To create Manifest, we need the cfg. The cfg is created by 
```
cfg, err := clientcmd.BuildConfigFromFlags(*MasterURL, *Kubeconfig)
```
so we need MasterURL and Kubeconfig. How can we access them in sharedmain?

markusthoemmes: You're right indeed, we can't reuse them cause they are captured in `MainWithConfig`.

I'd propose to make these flags package local (lowercase), use the plain `Main` function in our `cmd/manager/main.go` and reuse `sharedmain.GetConfig(*masterURL, *kubeconfig)` here locally. That function does all the defaulting needed by all controllers.

houshengbo: Is it sharedmain.GetConfig("", "")?

markusthoemmes: well, if you want to keep supporting the flags you gotta pass them in.

aliok: @markusthoemmes I made some changes.

pkg/reconciler/knativeserving/controller.go
```diff
@@ -39,8 +39,8 @@ const (
 
 var (
 	recursive  = flag.Bool("recursive", false, "If filename is a directory, process all manifests recursively")
-	MasterURL  = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
-	Kubeconfig = flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	masterURL  = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	kubeconfig = flag.String("kubeconfig", os.Getenv("KUBECONFIG"), "Path to a kubeconfig. Only required if out-of-cluster.")
```
markusthoemmes: The Getenv call is no longer needed as the `GetConfig` call below does that defaulting already.

aliok: Right! Thanks for the heads up.

aliok:
Squashed and ready for final review

aliok:
> for the nit. Feel free to unhold if you don't want to fix that.

I can't. I lack permissions :)

aliok:
oh man this failed :(

```
flag redefined: master
```

Can't redefine a flag. We either need to make the flags in sharedmain exposed or do some hack around reading the same flag again in our controller.

@markusthoemmes wdyt?

markusthoemmes:
Dang. For this PR I'd roll back the change to the "normal" `Main` function and use the `sharedmain.GetConfig` helper in `main.go` as well. That'll land your initial reason to open the PR in the first place.

We can then discuss what to do with the rest.

aliok:
> Cannot trigger testing until a trusted user reviews the PR and leaves an /ok-to-test message.

oh boy. 
cc @markusthoemmes

PR #294, opened 2020-02-12
Remove all the arguments for controller

houshengbo:
## Proposed Changes

* We will leverage sharedmain to get the configuration we need for operator use.

**Release Note**

```release-note

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -16,21 +16,10 @@ limitations under the License.
 package main
 
 import (
-	"flag"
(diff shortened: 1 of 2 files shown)


markusthoemmes:
This is a duplicate of #293, I think? Also: Should we really drop the master/kubeconfig options for manifestival usage? I feel like that'll cause weird behavior if somebody is using that and it's not consistently used between the "normal" clients and manifestival's.

/hold

For waiting for #293 to merge first.

PR #321, opened 2020-03-03
Add leader election

pmorie:
Fixes knative/pkg#1007

## Proposed Changes

- depends on knative/pkg#1019 for changes to `sharedmain`, which is used by:
  - serving-operator
- `/config` is changed so that it is simple to enable leader election, but it is not enabled by default

**Release Note**

```release-note
Add support for leader election
```


pmorie:
/hold

There are a couple things I want to run down before this merges

PR #322, opened 2020-03-04
Update the manifest of serving to 0.13.0

houshengbo:
Fixes #

## Proposed Changes

* This PR updates the manifest of serving to 0.13.0.

**Release Note**

```release-note

```


pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -290,5 +290,13 @@ func (r *Reconciler) deleteObsoleteResources(manifest *mf.Manifest, instance *se
 	if err := manifest.Client.Delete(resource); err != nil {
 		return err
 	}
+	// Remove Service activator from 0.12, since the new service is named activator-service
```
k4leung4: can we add an e2e test for this to verify that it is actually removed?

k4leung4: do we have a test that detected there were leftover objects?

houshengbo: Upgrade tests may be able to make sure these resources are removed.

k4leung4: can we update the upgrade test to ensure that after upgrading, there are no objects labeled for the previous version?

Adding logic that doesnt have corresponding tests makes me wonder if it is actually working or not.

Most objects have a release label, would it be safe to delete all objects from the old release?

jcrossley3: It's harmless for this to be in here, since it'll fail to delete anything silently, but it appears the name didn't change: https://github.com/knative/serving-operator/blob/326cbf7991bc801e631b2cef729be6ad7ca51991/cmd/manager/kodata/knative-serving/0.12.1.yaml#L784

houshengbo: @jcrossley3 @k4leung4 Was it just an illusion, or something wrong with my eyes?
I caught a Service named "activator", but cannot find it any more.

test/e2e-upgrade-tests.sh
```diff
@@ -155,6 +155,13 @@ install_serving_operator
 header "Running tests for Knative Serving Operator"
 failed=0
 
+# Verify with the bash script to make sure there is no resource with the label of the previous release.
+result="$(kubectl get all --no-headers -n ${TEST_NAMESPACE} -l serving.knative.dev/release=${LATEST_SERVING_RELEASE_VERSION} 2>/dev/null)"
+# If the ${result} is not empty, we fail the tests, because the resources from the previous release still exist.
+if [[ ! -z ${result} ]] ; then
+  fail_test "The resources with the label of previous release have not been removed."
```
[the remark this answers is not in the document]
houshengbo: @k4leung4 I print the result, if there is any resource left.

PR #326, opened 2020-03-05
Remove empty leader election config

pmorie:
#321 accidentally added an empty configmap, which we can't distribute via operatorhub currently. This PR removes it; when the configmap does not exist, leader election is disabled by default.

```release-note
NONE
```


pmorie:
/retest

looks like istio installation flaked

PR #330, opened 2020-03-06
Use ko's strict mode for better errors messages

mattmoor:


trshafer:
Looks like this works in serving: https://github.com/knative/serving/pull/7177

/lgtm
/approve

PR #333, opened 2020-03-08
Filter manifest to install/uninstall extensions such as serving-cert-manager, ns-cert and networking-istio

nak3:
This patch supports filter function to install/uninstall extensions
such as serving-cert-manager, networking-ns-cert and istio
ingress by the configuration. It solves following two problems.

#### problem 1

When `autoTLS` is enabled in config-network, we need
[namespace-wildcard-certs](https://github.com/knative/serving/tree/master/config/namespace-wildcard-certs) and [cert-manager](https://github.com/knative/serving/tree/master/config/cert-manager)
to be deployed. However current operator does not install them.

#### problem 2

When we want to change `ingress.class` to antoher Ingress solution
instead of Istio, current operator always deploys networking-istio and 
istio gateways.

To solve them, this patch filters manifest to install/uninstall
by the configuration.

e.g - when `autoTLS=Enabled`  `ingress.calss=kourier` is configured w/ this PR

```
$ kubectl get pod  -n knative-serving
NAME                                      READY   STATUS    RESTARTS   AGE
activator-869f6d4f9f-ll75v                1/1     Running   2          87m
autoscaler-78994c9fdf-c6kmn               1/1     Running   2          87m
autoscaler-hpa-66bcf9cdc-sxt72            1/1     Running   0          87m
controller-b94c5b667-97wkl                1/1     Running   2          87m
networking-certmanager-5bfcf45d79-kc958   1/1     Running   0          21m
networking-ns-cert-5447795455-wx9pj       1/1     Running   0          21m
webhook-7cdb467d79-4jrfq                  1/1     Running   2          87m
```

**Release Note**

```release-note
serving-cert-manager and networking-ns-cert are deployed when autoTLS is Enabled.
networking-istio is not deployed when ingress.class is not Istio.
```

/lint

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -195,8 +195,6 @@ func (r *Reconciler) filter(manifest *mf.Manifest, instance *servingv1alpha1.Kna
 	if len(labels) == 0 {
 		return *manifest, nil
 	}
-	// TODO: manifestival's Delete() prints error logs but never returns error actually, so this IsNoMatchError does not make sense.
-	// But leave this code here to show we cannot control "no matches for kind" erorr atm.
 	if err := manifest.Filter(byLabels(labels)).Filter(mf.NotCRDs).Delete(); err != nil && !meta.IsNoMatchError(err) {
 		return *manifest, err
 	}
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -47,6 +47,12 @@ const (
 	deletionChange = "deletion"
 )
 
+var (
+	autoTlsLabels = map[string]string{
+		"networking.knative.dev/certificate-provider":          "cert-manager",
+		"networking.knative.dev/wildcard-certificate-provider": "nscert"}
+)
+
 // Reconciler implements controller.Reconciler for Knativeserving resources.
 type Reconciler struct {
 	*reconciler.Base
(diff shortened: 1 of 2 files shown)


pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -164,6 +176,33 @@ func (r *Reconciler) transform(instance *servingv1alpha1.KnativeServing) (mf.Man
...
+	}
+	if len(labels) == 0 {
+		return *manifest, nil
+	}
+	// TODO: manifestival's Delete() prints error logs but never returns error actually, so this IsNoMatchError does not make sense.
+	// But leave this code here to show we cannot control "no matches for kind" erorr atm.
+	if err := manifest.Filter(byLabels(labels)).Filter(mf.NotCRDs).Delete(); err != nil && !meta.IsNoMatchError(err) {
+		return *manifest, err
+	}
+	return manifest.Filter(byNoLabels(labels)), nil
```
jcrossley3: FYI: `byNoLabels(labels)` is equivalent to `mf.Complement(byLabels(labels))`

In v0.3.0, I renamed `mf.Complement` to `mf.None`

There is already a `mf.ByLabel` filter, too, but it only takes a single key/value. It'd be nice to have a `mf.ByLabels` alternative that takes the map.

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -164,6 +176,33 @@ func (r *Reconciler) transform(instance *servingv1alpha1.KnativeServing) (mf.Man
...
+	if !isAutoTLSEnabled(manifest, instance) {
+		r.Logger.Debug("Removing autoTLS extention resources")
+		for k, v := range autoTLSLabels {
+			labels[k] = v
+		}
+	}
+	if len(labels) == 0 {
+		return *manifest, nil
+	}
+	// TODO: manifestival's Delete() prints error logs but never returns error actually, so this IsNoMatchError does not make sense.
```
jcrossley3: I think I consider this a bug. I'll write up an issue and fix.

nak3:
/test pull-knative-serving-operator-upgrade-tests

`TestProbe` failed. I think it is flaky.

nak3:
I'm not sure if serverless-opeator will support cert-manager, ns-cert and other ingress solutions. So I'm closing this rather than keep rebasing it.

PR #337, opened 2020-03-10
Add support for HA control plane

pmorie:
Fixes #320 

## Proposed Changes

* Adds `KnativeServing.spec.highAvailability`

**Release Note**

```release-note
Initial support for highly available control plane
```

PR #338, opened 2020-03-11
Bump to manifestival 0.4.0

jcrossley3:
Fixes a pretty gnarly bug where transformers using `Scheme.Convert` don't work.

This was exposed by #337


chizhg:
@jcrossley3 can you type `@googlebot I consent.` here? This seems to be a googlebot's problem.

PR #355, opened 2020-03-12
Update yaml

houshengbo:
Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```


k4leung4:
can we write simple script for updating the yaml file? as it is effectively a big blob, its hard to review, and would be nice if people can reproduce the yaml update.

for the version, what is the format for the version field do we want? feels like we flip flop between the knative version and a date-hash string?

PR #359, opened 2020-03-12
Delete (Cluster)Role(Bindings) as a final cleanup step.

Cynocracy:
This allows us to take advantage of the permissions granted to cluster admins
when performing cleanup. If the approach in https://github.com/knative/serving-operator/pull/291
is approved, this will become necessary. Additionally, it makes sense to remove roles as the final
step to allow human Operators to modify any resources they may have permissions on as a result of the Knative installation (that is, we should not remove any access until we are 'almost done' cleaning up).

/lint

## Proposed Changes

* Delete roles and rolebindings as a final cleanup step.

**Release Note**

```release-note
Delete roles and rolebindings as a final cleanup step.
```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -251,7 +251,7 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
 		if err := r.config.Filter(mf.ByKind("Deployment")).Delete(); err != nil {
 			return err
 		}
-		if err := r.config.Filter(mf.All(mf.NoCRDs, mf.None(RBAC))).Delete(); err != nil {
+		if err := r.config.Filter(mf.NoCRDs, mf.None(RBAC)).Delete(); err != nil {
 			return err
 		}
 		// Delete Roles last, as they may be useful for human operators to clean up.
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -255,7 +255,7 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
 			return err
 		}
 		// Delete Roles last, as they may be granting us permissions that are necessary.
-	        if err := r.config.Filter(RbacResources).Delete(); err != nil {
+		if err := r.config.Filter(RbacResources).Delete(); err != nil {
 			return err
 		}
 	}
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -246,11 +246,16 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
 	if len(instance.GetFinalizers()) == 0 || instance.GetFinalizers()[0] != finalizerName {
 		return nil
 	}
+	var RbacResources = mf.Any(mf.ByKind("Role"), mf.ByKind("ClusterRole"), mf.ByKind("RoleBinding"), mf.ByKind("ClusterRoleBinding"))
 	if len(r.servings) == 0 {
 		if err := r.config.Filter(mf.ByKind("Deployment")).Delete(); err != nil {
 			return err
 		}
-		if err := r.config.Filter(mf.NoCRDs).Delete(); err != nil {
+		if err := r.config.Filter(mf.All(mf.NoCRDs, mf.None(RbacResources))).Delete(); err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


k4leung4:
maybe add a unit test?

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -246,11 +246,16 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
 	if len(instance.GetFinalizers()) == 0 || instance.GetFinalizers()[0] != finalizerName {
 		return nil
 	}
+	var RBAC = mf.Any(mf.ByKind("Role"), mf.ByKind("ClusterRole"), mf.ByKind("RoleBinding"), mf.ByKind("ClusterRoleBinding"))
 	if len(r.servings) == 0 {
 		if err := r.config.Filter(mf.ByKind("Deployment")).Delete(); err != nil {
 			return err
 		}
-		if err := r.config.Filter(mf.NoCRDs).Delete(); err != nil {
+		if err := r.config.Filter(mf.All(mf.NoCRDs, mf.None(RBAC))).Delete(); err != nil {
```
jcrossley3: The `mf.All` call is redundant. You feel it adds clarity?

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -246,11 +246,16 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
...
+			return err
+		}
+		// Delete Roles last, as they may be useful for human operators to clean up.
+		if err := r.config.Filter(RBAC).Delete(); err != nil {
```
jcrossley3: Why exactly does the order matter? It's not as if they're going to be around long enough for a human to benefit from them are they?

pkg/reconciler/knativeserving/knativeserving_controller.go
Cynocracy: Nope, didn't notice the redundancy, I'll remove.

pkg/reconciler/knativeserving/knativeserving_controller.go
Cynocracy: In the happy case where the Operator is able to successfully clean up state, yes. In the case where the Operator fails for some reason (user misconfigured RBAC policies, etc) it's better to keep roles like knative-serving-admin ( https://github.com/knative/serving/blob/master/config/core/200-serviceaccount.yaml#L26 ) around so that the human might have the ability to clean things up.

The other reason is https://github.com/knative/serving-operator/pull/291
In that change, the Operator 'bootstrap's itself into having permissions on all resources, by having 'escalate' on all roles, and being automatically granted all clusterroles with 'serving.knative.dev/release', the Operator is able to grant itself any necessary permissions without needing to know exactly what they entail. If the Operator deletes the clusterroles that are allowing it admin access, it may lose the ability to delete resources it still needs to clean up.

jcrossley3: Gotcha, thanks for the explanation!

Cynocracy: No problem :) Thanks for the review!

Cynocracy:
@googlebot I consent

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -246,11 +246,16 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
...
+	var RBAC = mf.Any(mf.ByKind("Role"), mf.ByKind("ClusterRole"), mf.ByKind("RoleBinding"), mf.ByKind("ClusterRoleBinding"))
 	if len(r.servings) == 0 {
 		if err := r.config.Filter(mf.ByKind("Deployment")).Delete(); err != nil {
 			return err
 		}
-		if err := r.config.Filter(mf.NoCRDs).Delete(); err != nil {
+		if err := r.config.Filter(mf.NoCRDs, mf.None(RBAC)).Delete(); err != nil {
+			return err
+		}
+		// Delete Roles last, as they may be useful for human operators to clean up.
```
k4leung4: can we add a unit test to verify this behavior

Cynocracy:
@k4leung4 re: unit test, I'm not sure I'd like to expand the scope of this PR to include one.

Perhaps I can work on one in a separate issue/change?

k4leung4:
@Cynocracy works for me.
just want to make sure we verify behavior changes as we dont really know when it stops working.

/approve
/lgtm

Cynocracy:
Gotcha, there are integration tests here:
https://github.com/knative/serving-operator/blob/master/test/resources/verify.go#L190

Filed https://github.com/knative/serving-operator/issues/361 to add unit tests.

Issue #329, opened 2020-03-06
Allow CR to handle manifest patches


kramvan1:
**Problem**

The custom resource (CR) for the knative operators should allow for tweaking all the resources deployed at the Kuberneters spec level. For example, allow for Affinity, Envs, HPA, Image pull policy, Annotations, Replicas, probes (readiness, liveness, startup), LifeCycle, Priority class, Tolerations, Request sizes...

For example, there are cases where the default deployment needs to be tweaked for a specific usage.
In one case we needed to add a `Lifecycle preStop` to a deployment to allow it more time to shutdown.

An example of how Istio handled this in the CR looked like this:

```
    ingressGateways:
    - name: istio-ingressgateway
      enabled: true
      k8s:
        overlays:
        - kind: Deployment
          name: istio-ingressgateway
          patches:
          - path: spec.template.spec.containers.[name:istio-proxy].lifecycle
            value:
              preStop:
                exec:
                  command: ["sleep", "25"]
```
Where ingressGateways in the high level CR component and istio-ingressgateway is the section of the manifest we are interested in patching.

The patch is straight forward, just a basic spec path and values to patch in there.

Which persona is this feature for?

integrator

**Exit Criteria**

See the deployment yaml output changed to include the patched values.

**Time Estimate (optional):**

not sure.

**Additional context (optional)**

Refs:
https://istio.io/docs/setup/install/standalone-operator/
https://istio.io/docs/setup/install/istioctl/#customize-kubernetes-settings

aliok:
Related: https://github.com/knative/serving-operator/issues/302

houshengbo:
We would like implement on a use case based approach, instead of offering one powerful way to override everything.
Another related topic: https://github.com/knative-sandbox/operator/issues/5

Issue #302, opened 2020-02-18
Support for advanced k8s resource overlays for configuration


Fabian-K:
**Problem**
The operator currently supports limited customization options via the crd (e.g. ingress selectors, registry credentials etc). For often-used customizations, It´s great if they are part of the crd. There are however many more customizations that sometimes require adjustment ( e.g. #160 #161 ). Adding all possibilities to the crd might not be an good idea.

The istio-operator supports so called "advanced k8s resource overlays" https://github.com/istio/operator#advanced-k8s-resource-overlays. To me, this looks like a great escape hatch to enable advanced customization of all values that are not yet (and probably will never be) exposed via the crd.

PR #362, opened 2020-03-14
Resources override

### LINKED ISSUE #161: support to tweak resource limits and requests for knative deployments
limits and requests are defined like https://github.com/knative/serving-operator/blob/master/cmd/manager/kodata/knative-serving/0.8.0.yaml#L752-L758, but they are not configurable.

E.g - If we changed resources.limits by `kubectl edit deployment autoscaler -n knative-serving`, it is reverted with same reason of https://github.com/knative/serving-operator/issues/160.

jcrossley3:
Fixes #161 
Related #302 #329 

## Proposed Changes

* Introduce `resources` field in the `KnativeServing` spec, which is a list of `ResourceRequirementsOverrides`, enabling users to override the resources consumed by any of the containers listed in the embedded manifest.  

This exposes a "knob" in the `KnativeServing` CR that would enable users to workaround issues like knative/serving#7195

```
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeServing
metadata:
  name: knative-serving
spec:
  resources:
  - container: webhook
    requests:
      cpu: 20m
      memory: 20Mi
    limits:
      cpu: 400m
      memory: 800Mi
```

**Release Note**

```release-note
Resource limits for any knative deployment containers can now be configured in the `KnativeServing` custom resource using the new `resources` field.
```


markusthoemmes:
> By default, no resource limits will be specified for any knative deployment containers.

I think this bit should be discussed and implemented upstream first 🤔. It doesn't feel right to have the defaults of both installation modes differ that vastly.

pkg/reconciler/knativeserving/common/resources.go
```diff
@@ -0,0 +1,50 @@
...
+func ResourceRequirementsTransform(instance *servingv1alpha1.KnativeServing, log *zap.SugaredLogger) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() == "Deployment" {
+			deployment := &appsv1.Deployment{}
+			if err := scheme.Scheme.Convert(u, deployment, nil); err != nil {
+				return err
+			}
+			containers := deployment.Spec.Template.Spec.Containers
+			for i := range containers {
+				containers[i].Resources = instance.Spec.Resources[containers[i].Name]
```
markusthoemmes: This will not only drop limits if empty, but also `requests`, which are important for scheduling decisions. I'm wondering if we want to only override those if explicitly set and keep the default requests in the manifest.

jcrossley3: Easy enough. Won't change defaults, but does expose knobs for tweaking.

pkg/reconciler/knativeserving/common/resources_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	for _, ks := range tests {
+		t.Run(ks.Name, func(t *testing.T) {
+			runResourceRequirementsTransformTest(t, &ks)
+		})
+	}
+}
+
+func runResourceRequirementsTransformTest(t *testing.T, ks *servingv1alpha1.KnativeServing) {
+	manifest, err := mf.NewManifest("testdata/manifest.yaml")
```
houshengbo: I am wondering if we can drop the file at pkg/reconciler/knativeserving/common/testdata/manifest.yaml, and reuse the one at test/config, which is an existing Symbolic link to cmd/manager/kodata/knative-serving.

To use that link: we can try 
```
_, filename, _, _ := runtime.Caller(0)
manifest, err := mf.NewManifest(filepath.Join((filepath.Dir(filename)+"/../../.."), "test/config/"))
```

jcrossley3: I expect the upstream resource config to change over time, so I'd prefer not to have to update the test whenever we update the manifest. I just needed some test fixtures, and I was too lazy to create them in Go -- besides yaml's much easier to maintain.

jcrossley3:
I'm going to change the logic to not delete the upstream config, by default, so...
/hold

pkg/reconciler/knativeserving/common/resources.go
pmorie: +1, I would expect a field by field merge here so that empty values in the operator CR don't trash field that are set in the manifest but not the operator CR

pkg/apis/serving/v1alpha1/knativeserving_types.go
```diff
@@ -105,6 +105,10 @@ type KnativeServingSpec struct {
 	// Allows specification of HA control plane
 	// +optional
 	HighAvailability *HighAvailability `json:"high-availability,omitempty"`
+
+	// Override resource requirements; keyed by container name
+	// +optional
+	Resources map[string]corev1.ResourceRequirements `json:"resources,omitempty"`
```
pmorie: The API conventions discourage maps for fields like this: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

jcrossley3: Good call. A struct with a `container` field is much clearer than a string key.

pkg/reconciler/knativeserving/common/resources.go
jcrossley3: Done

pkg/apis/serving/v1alpha1/knativeserving_types.go
jcrossley3: Done.

PR #369, opened 2020-03-19
Update the manifest to the latest nightly build

houshengbo:
Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```

### DIFF
--- a/cmd/manager/kodata/knative-serving/serving.yaml
+++ b/cmd/manager/kodata/knative-serving/serving.yaml
@@ -26,10 +26,10 @@ spec:
     plural: images
     singular: image
     categories:
-    - knative-internal
-    - caching
+      - knative-internal
+      - caching
     shortNames:
-    - img
+      - img
   scope: Namespaced
   subresources:
     status: {}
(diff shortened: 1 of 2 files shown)


k4leung4:
How often do we have to update the manifest to the latest nightly?
Should we investigate into automation if this should be done frequently?

houshengbo:
@k4leung4 I plan to do it on a weekly basis, but there are several issues:
1. Inconsistency of labels used in one nightly built yaml:
If we download the latest serving yaml here https://storage.googleapis.com/knative-nightly/serving/latest/serving.yaml. I see the deployment
name: networking-istio
has the label
serving.knative.dev/release: "v20200316-516fac9"
, but the deployment
name: autoscaler-hpa
has the label
serving.knative.dev/release: "v20200319-2f826455d"
Why do we have two different labels?
If we do it automatically, how to get the version number?

2. How to change the version number in version.go, if we do the update automatically?

k4leung4:
1. the release labels are different because net-istio is built from knative.dev/net-istio and vendored, https://github.com/knative/serving/pull/7266/files

2. we only use release versions for releases right? so maybe have a daily job that updates with the nightly build. for releases, we manually do update with the release build, cut the release branch, and the daily job will update it back to the nightly build?

houshengbo:
The issue is with the unit tests TestManifestVersionSame.
If the version does not match the version number in manifest, they fail.

PR #372, opened 2020-03-23
Print log message when start deleting resources

nak3:
## Proposed Changes

This patch makes a tiny change to add log message when operator
deletes resources.

Currently there are no logs so it is difficult to debug when resources
are not deleted.

After this patch:
```
{"level":"info","ts":"2020-03-23T05:36:28.296Z","logger":"serving-operator.knativeserving-controller","caller":"knativeserving/knativeserving_controller.go:249","msg":"Deleting resources","commit":"d558b3a","knative.dev/controller":"knativeserving-controller"}
```

/lint

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -246,6 +246,7 @@ func (r *Reconciler) delete(instance *servingv1alpha1.KnativeServing) error {
 	if len(instance.GetFinalizers()) == 0 || instance.GetFinalizers()[0] != finalizerName {
 		return nil
 	}
+	r.Logger.Info("Deleting resources")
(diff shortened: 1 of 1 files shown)


nak3:
Some more context.

I am looking into the issue about the resource deletion issue. The CI produced following [error](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/185/pull-ci-openshift-knative-serverless-operator-master-4.3-e2e-aws-ocp-43/559):

```
    --- FAIL: TestKnativeServingDeployment/delete (11.26s)
        verify.go:216: The ClusterRole knative-serving-admin failed to be deleted: <nil>
```

But no clue in the logs and no "Deleting" message from manifestival. So, this patch prints the logs to see if operator triggered the delete func or not.

jcrossley3:
> But no clue in the logs and no "Deleting" message from manifestival.

Just an FYI, but manifestival will typically only delete those resources in the manifest the garbage collector fails to reap when the "owning" KS resource is deleted. There is a race condition there, obviously, which shouldn't be a problem but will explain why you might not see any "Deleting" message.

PR #374, opened 2020-03-25
Install from the manifest in a strict(er) order.

Cynocracy:
This order, Roles -> RoleBindings -> The Rest, prevents the operand from
needing to structure their manifest such that this ordering exists, while
allowing the Operator to use the bootstrapping mechanism described in
https://github.com/knative/serving-operator/pull/291 to 'escalate' itself
into management of the Knative Serving installation without a *** clusterrole.

## Proposed Changes

* Install roles, then rolebindings, then the remainder of the manifest.

**Release Note**

```release-note
NONE

```

### DIFF
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -207,7 +207,7 @@ func (r *Reconciler) install(manifest *mf.Manifest, instance *servingv1alpha1.Kn
 		instance.Status.MarkInstallFailed(err.Error())
 		return err
 	}
-	if err := manifest.Filter(mf.None(mf.Any(role, rolebinding))).Apply(); err != nil {
+	if err := manifest.Apply(); err != nil {
 		instance.Status.MarkInstallFailed(err.Error())
 		return err
 	}
--- a/pkg/reconciler/knativeserving/knativeserving_controller.go
+++ b/pkg/reconciler/knativeserving/knativeserving_controller.go
@@ -47,6 +47,11 @@ const (
 	deletionChange = "deletion"
 )
 
+var (
+	role mf.Predicate = mf.Any(mf.ByKind("ClusterRole"), mf.ByKind("Role"))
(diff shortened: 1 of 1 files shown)


pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -47,6 +47,11 @@ const (
 	deletionChange = "deletion"
 )
 
+var (
+	role mf.Predicate = mf.Any(mf.ByKind("ClusterRole"), mf.ByKind("Role"))
```
mattmoor: Format Go code:
```suggestion
	role        mf.Predicate = mf.Any(mf.ByKind("ClusterRole"), mf.ByKind("Role"))
```

Cynocracy: Sorry mattbot :) fixed.

Cynocracy:
Will probably need to find a different assignee :(

pkg/reconciler/knativeserving/knativeserving_controller.go
```diff
@@ -191,7 +196,18 @@ func (r *Reconciler) initStatus(_ *mf.Manifest, instance *servingv1alpha1.Knativ
...
+	// (Cluster)RoleBindings, then the rest of the manifest.
+	if err := manifest.Filter(role).Apply(); err != nil {
+		instance.Status.MarkInstallFailed(err.Error())
+		return err
+	}
+	if err := manifest.Filter(rolebinding).Apply(); err != nil {
+		instance.Status.MarkInstallFailed(err.Error())
+		return err
+	}
+	if err := manifest.Filter(mf.None(mf.Any(role, rolebinding))).Apply(); err != nil {
```
jcrossley3: This is fine, but technically you could just `manifest.Apply()` without the filtering. Any resources already applied will be no-ops.

Cynocracy: I like the simplicity, updated :)

Cynocracy:
Thanks for the review :)

PR #381, opened 2020-03-30
Update knative/pkg from 0.13 to master

houshengbo:
Fixes #

## Proposed Changes

*
*
*

**Release Note**

```release-note

```


Cynocracy:
Is it safe to point at master instead of a particular release?

houshengbo:
@Cynocracy We can pin to master before a certain release cut, in order to catch all the updates in dependency. When the package release 0.14 for example, we will pin to that release branch to prepare our 0.14 release.

Cynocracy:
Thanks, that makes sense!

/lgtm

PR #387, opened 2020-04-01
Fix the retry issue with deployment verification

houshengbo:
Fixes #

## Proposed Changes

* This PR fixed the broken retry mechanism, checking whether the deployments reach the desired state. It used to return `false, err`, which would run the check only once, but we need to return `false, nil`, till we reach the desired state. This PR made the correct changes for the retry mechanism.

**Release Note**

```release-note

```


aliok:
@houshengbo 
Can you give some details about this PR?
I think the code changes look fine, but just want to understand what the original problem was.

houshengbo:
@aliok I added in the comments of the PR.

aliok:
/lgtm

makes perfect sense :+1:

PR #394, opened 2020-04-07
Respect HA Replicas setting in HPAs.

markusthoemmes:
## Proposed Changes

If the operator has a HighAvailabiliy configuration and the replicas are set > 1, we probably want to adjust all HPAs to have at least as many minReplicas as specified in that HA config. An HPA being present is enough of a signal for us to assume that the component supports a HA configuration.

**Release Note**

```release-note
Adjusts all HPAs to have at least the number of replicas specified by a potential HighAvailability configuration.
```

### DIFF
--- a/pkg/reconciler/knativeserving/common/ha.go
+++ b/pkg/reconciler/knativeserving/common/ha.go
@@ -62,9 +62,26 @@ func HighAvailabilityTransform(instance *servingv1alpha1.KnativeServing, log *za
 			}
 		}
 
+		replicas := int64(instance.Spec.HighAvailability.Replicas)
+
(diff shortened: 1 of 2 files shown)


pmorie:
in practice, based on the current resources in serving `config` directory, i expect this initially to set `spec.minReplicas` of the activator HPA. Is that accurate?

pmorie:
Discussed w/ @markusthoemmes offline that the behavior is what I expect.

/lgtm

houshengbo:
@markusthoemmes Would you mind contributing this PR to the new repo at https://github.com/knative-sandbox/operator, since we have set up the consolidated oprator.

PR #407, opened 2020-04-21
Pass config from main to the controller.

markusthoemmes:
## Proposed Changes

To align with https://github.com/knative-sandbox/operator/pull/21 and to be able to use this controller as a "library" (to compile multiple operators into one binary), this moves the flag definition to the main.go file and passes the built config through.

**Release Note**

```release-note
NONE
```

/assign @jcrossley3 @houshengbo

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -24,12 +24,17 @@ import (
 	"knative.dev/serving-operator/pkg/reconciler/knativeserving"
 )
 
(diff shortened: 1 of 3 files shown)


aliok:
/approve
/lgtm

Makes a lot of sense.
I know we're not really going to use this operator's pkg as a library, but aligning with the new operator is nice.

PR #414, opened 2020-05-18
Fix the retry issue with deployment verification (#387)

houshengbo:
The retry mechanism in the test cases ONLY get sand verifies the resources ONCE, since it always returns an error in the retry logic. To make the retry really work, we need to return nil for next retry before timeout.

## Proposed Changes

* This PR is a cherry-pick of acb5e5d23b01d8afd82a4f4c4f0c73bc9847f90c, resolving the issue of ONLY verifying the resource ONCE.

**Release Note**

```release-note

```


markusthoemmes:
Can you please provide a description of which issue is fixed here? I have no clue what the "retry issue with deployment verification" is to be honest.

houshengbo:
@markusthoemmes I had this fix in release 0.14.
It was my fault of only trying ONLY ONCE to verify the resources of the knative deployments.
I suspected the failure of https://github.com/knative/serving-operator/pull/413, was introduced by the defected retrying.

houshengbo:

